PR #2, opened 2017-03-24
Search lc

jduss4:
adds more lewis and clark functionality to search by way of the summary boxes at the top of search results


app/helpers/orchid/application_helper.rb
```diff
@@ -4,4 +4,11 @@ def copy_params
     return params.to_unsafe_h
   end
 
+  def clear_search_text
+    new_params = copy_params
+    # TODO also remove qfield from here if that is implemented
```
techgique: Since we decided against using `qfield`, I suppose this line can go?

PR #3, opened 2017-03-27
Generator

jduss4:
Quickly prepares a new rails app for use with orchid!

There's more that we can do here, like copying in directory structures, helper file templates, etc, but I think this is all we need to get started!

### DIFF
--- a/lib/generators/setup_generator.rb
+++ b/lib/generators/setup_generator.rb
@@ -23,11 +23,12 @@ class SetupGenerator < Rails::Generators::Base
     msgs << facets
     msgs << stylesheet
 
-    puts msgs.join("\n")
-
     Bundler.with_clean_env do
       run "bundle install"
     end
+
+    puts msgs.compact.join("\n")
+
   end
 
   private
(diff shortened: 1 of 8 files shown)


lib/generators/setup_generator.rb
```diff
@@ -0,0 +1,94 @@
...
+    else
+      puts "Found existing file in facets.rb, skipping"
+    end
+  end
+
+  def gems
+    # remove the previous gem from Gemfile
+    gsub_file "#{@app_dir}/Gemfile", /^(?!#\s)gem\s["']api_bridge["'].*$/, ""
+    # install the correct version of the gem
+    gem "api_bridge", git: "https://github.com/CDRH/api_bridge", version: "0.0.1"
```
jduss4: I would like to, yes.  I'm guessing that it could be somewhat similar to the `config.rb` style setup of a normal rails app?  Maybe I'll work on that over lunch.  This is how Blacklight does it, though I don't know how much I care for it:

https://github.com/projectblacklight/blacklight/blob/master/lib/blacklight/version.rb

lib/generators/setup_generator.rb
```diff
@@ -0,0 +1,94 @@
...
+    # http://guides.rubyonrails.org/generators.html#initializer
+    FileUtils.cp("#{@this_dir}/templates/config.rb", "#{@app_dir}/config/initializers/config.rb")
+
+    # example config file
+    FileUtils.cp("#{@this_dir}/templates/config.yml", "#{@app_dir}/config/config.example.yml")
+
+    # config file
+    # does not overwrite existing config file
+    if File.exist?("#{@app_dir}/config/config.yml")
+      return "Verify that your config/config.yml file matches the options in config/config.example.yml".cyan
```
jduss4: I think that they won't be identical, because they should be customizing the config file to have their server, so it won't match the config file being copied in from here.  It will only copy if there isn't a file there.  I'll go through and make sure that exist vs non-exist is standardized

PR #4, opened 2017-04-17
fixes generator to work with brand new app

jduss4:
removes any "updating" stuff in favor of a straight setup script
has root path of both apps instead of one root, one generator directory
adds bootstrap gem, styling scss, etc


lib/generators/setup_generator.rb
```diff
@@ -34,64 +37,64 @@ def setup_files
...
-    else
-      FileUtils.cp("#{@this_dir}/templates/config.yml", "#{@app_dir}/config/config.yml")
-      return "Customize your app in config/config.yml".green
-    end
+    FileUtils.cp("#{@this_app}/lib/generators/templates/config.yml", "#{@new_app}/config/config.yml")
+    return "Customize your app in config/config.yml".green
+  end
+
+  def copy_initializer
+    # NOTE: This could be done with the "initializer" method instead
```
techgique: I scribbled out what it would take:
```ruby
file = File.open("#{@this_app}/lib/generators/templates/config.rb", "rb")
initializer("config.rb", file.read)
file.close
```
I think the existing code is a much briefer, more straightforward approach 👍 
The method approach just seems quicker for dropping in static strings or new secrets etc.

app/controllers/items_controller.rb
```diff
@@ -5,6 +5,7 @@ def browse
   end
 
   def index
+    @extra_js = [javascript_include_tag("search")]
```
jduss4: if we wanted to avoid the `html_safe` business, presumably we could run the `javascript_include_tag` on any strings passed in from this array below.

```ruby
 <% if !@extra_js.blank? %>
  <% Array.wrap(@extra_js).each do |js|
    javascript_include_tag(js)
  end %>
<% end %>

or something like that?

File: app/views/layouts/application.html.erb [hunk 3]
```diff
@@ -1,26 +1,45 @@
...
+  <%#= favicon_link_tag "favicon.png", rel: "icon", sizes: "192x192", type: nil %>
+  <%#= favicon_link_tag "apple-touch-icon.png", rel: "apple-touch-icon", sizes: "180x180", type: nil %>
+  <!-- <meta name="msapplication-config" content="<%= browserconfig_path format: :xml %>"> -->
+  <%#= favicon_link_tag "favicon.ico", rel: "icon", type: nil %>
 
   <!-- Stylesheets -->
   <%= stylesheet_link_tag 'application', media: 'all' %>
+  <% if !@extra_css.blank? %>
+    <% Array.wrap(@extra_css).each do |css|
+      concat "  #{css}".html_safe
```
jduss4: same here as with the javascript, I feel like it's a bit cleaner to have the `stylesheet_link_tag` here instead of in the controller

app/views/layouts/application.html.erb
```diff
@@ -1,26 +1,45 @@
...
+    end %>
+  <% end %>
 
   <body>
     <div class="header">
       <div class="container">
         <%= link_to home_path do %>
           <div class="site_title">
             <h1><%= CONFIG["project_name"] %></h1>
-            <h2>Subtitle for CDRH Elasticsearch API Sites</h2>
```
jduss4: should we include this as a config setting, too?

app/views/layouts/application.html.erb
```diff
@@ -55,7 +74,7 @@
         <div class="footer_info">
           <p>Created by the <a href="http://cdrh.unl.edu">Center for Digital Research in the Humanities</a>.</p>
           <div class="footer_logo">
-            <a href="http://www.unl.edu"><%# image_tag 'unl_black.png' %></a>
+            <a href="http://www.unl.edu"><%= image_tag 'unl_black.png' %></a>
```
jduss4: unfortunately, this is going to fail when a new app starts.  The fix for it is in a different branch:  https://github.com/CDRH/orchid/commit/6ec1cea66b0f03a20c8591b11d7c4b7a36e22a3e

I recommend that we simply leave it commented for now to avoid the app failing on starting until the other branch is merged

lib/generators/setup_generator.rb
```diff
@@ -56,9 +55,14 @@ def facets
     return "Customize your facets in app/models/facets.rb".green
   end
 
-  def gems
+  def footer_logo
```
jduss4: oh, lol, now I get why it was above :)

Which solution do you prefer?  Adding the favicon and footer logo to the new app like this?  Or adding them to the precompile step?

lib/generators/setup_generator.rb
```diff
@@ -74,27 +78,30 @@ def gitignore
...
   def scripts
-    # copy over application.js
     FileUtils.cp("#{@this_app}/app/assets/javascripts/application.js", "#{@new_app}/app/assets/javascripts/application.js")
-    return "Add custom site-wide javascript files to app/assets/javascripts/\nAll .js files located there are included on every page".green
+    FileUtils.cp("#{@this_app}/app/assets/javascripts/search.js", "#{@new_app}/app/assets/javascripts/search.js")
+
+    # Create directory for auto-included app-wide JavaScript
+    FileUtils.mkdir("#{@new_app}/app/assets/javascripts/global")
+
+    return "Add app-wide JavaScript to app/assets/javascripts/global/\nAll .js file contents there are served with every page\n\nView-specific JS is to be added to @extra_js instance variable, e.g.:\n  @extra_js = [javascript_include_tag(\"search\")]".green
```
jduss4: only commenting here in case the js_include_tag gets moved to the views

lib/generators/setup_generator.rb
```diff
@@ -74,27 +78,30 @@ def gitignore
...
+    # Bootstrap variable overrides
     FileUtils.cp("#{@this_app}/app/assets/stylesheets/cdrh-bootstrap-variables.scss", "#{@new_app}/app/assets/stylesheets/cdrh-bootstrap-variables.scss")
 
-    # sub application.css for scss with variables
+    # Must use application.scss for mixins and variables
     FileUtils.rm("#{@new_app}/app/assets/stylesheets/application.css")
     FileUtils.cp("#{@this_app}/app/assets/stylesheets/application.scss", "#{@new_app}/app/assets/stylesheets/application.scss")
 
-    return "Customize your stylesheet in app/assets/stylseheets/cdrh-bootstrap-variables.scss".green
+    return "Customize Bootstrap in app/assets/stylseheets/cdrh-bootstrap-variables.scss\nApplication-wide styling to be added via application.scss\n\nView-specific styling to be added to @extra_css instance variable, e.g.:\n  @extra_css = [stylesheet_link_tag(\"leaflet\")]".green
```
jduss4: same comment to keep in mind

app/controllers/items_controller.rb
techgique: Sounds good to me! I just copied how we did it in L&C so we should probably change it there too. Would you mind updating it over there and I'll fix it in orchid?

app/views/layouts/application.html.erb
techgique: Yeah, I'm down with that. Just didn't want a string in there we'd have to hunt down as it was.

app/controllers/items_controller.rb
jduss4: No problem!  I must have completely missed that in L&C, whoops!

app/views/layouts/application.html.erb
techgique: Will leave this in and add the favicon from the other branch after this is merged

lib/generators/setup_generator.rb
techgique: Will update these as well 👍

PR #6, opened 2017-04-26
partials

jduss4:
Splits mostly the `<head>` element into separate templates, moves search partials into items view directory


app/views/layouts/application.html.erb
```diff
@@ -1,95 +1,11 @@
...
-              <li><%= link_to "About", about_path %></li>
-            </ul>
-          </div>
-        </div>
-      </div><!-- /navbar -->
-    </div><!-- /header -->
-
-    <div class="container" id="wrapper">
-      <%= yield %>
+    <div class="main">
```
techgique: @karindalziel 
Is this related to interest in a page-wide class named for the controller or site section? o.O
https://github.com/CDRH/api/issues/50

techgique:
I'll review and sync up the little differences between the static assets code between this repo and L&C later tonight and test things out. Looks like we should be good to go otherwise.

app/views/layouts/application.html.erb
jduss4: I don't think so. I believe that @karindalziel requested this div also, however.  Can you weigh in, @karindalziel ?

karindalziel: The page wide class was supposed to be on the html or body tag, not on this div. (And I'm hoping can be based on the route, rather than the controller - so things at /about will have body class="site_about" or something like that).  

This div: in the past when I don't have full control over the HTML, I sometimes with there was another div inside body but outside header body and footer. I figured it's safest to add it now in case I or someone else needs it for styling later. 

Note: this should NOT have the class container, as that's a reserved bootstrap class. I would just leave the ID as wrapper and get rid of the class.

PR #7, opened 2017-05-03
Sync Rails conditional asset templating

techgique:
Remove use of concat and use ERB output

Add "media: nil" option to stylesheet tag,
as Rails defaults to setting `media="screen"`
but browsers default to treating as `media="all"` without it

Add output of  comment if conditional assets loaded

<%= stylesheet_link_tag ... %> and <%= javascript_include_tag ...%>
are indented awkwardly in the template source
to produce well-indented markup in page source

### DIFF
--- a/app/views/layouts/head/_scripts.html.erb
+++ b/app/views/layouts/head/_scripts.html.erb
@@ -1,12 +1,17 @@
-<!-- Scripts -->
-<%= javascript_include_tag 'application' %>
-<% if @ext_js.present?
-  Array.wrap(@ext_js).each do |js|
-    concat "  #{javascript_include_tag js}\n".html_safe
-  end
-end %>
-<% if @inline_js.present?
-  Array.wrap(@inline_js).each do |js|
-    concat "  <script>\n#{js}\n  </script>\n".html_safe
-  end
-end %>
+  <!-- Scripts -->
+  <%= javascript_include_tag 'application' %>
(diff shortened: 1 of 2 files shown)


app/views/layouts/head/_scripts.html.erb
```diff
@@ -1,12 +1,17 @@
...
-end %>
-<% if @inline_js.present?
-  Array.wrap(@inline_js).each do |js|
-    concat "  <script>\n#{js}\n  </script>\n".html_safe
-  end
-end %>
+  <!-- Scripts -->
+  <%= javascript_include_tag 'application' %>
+  <% if @ext_js.present? || @inline_js.present? %>
+  <!-- Conditional -->
```
jduss4: You pointed this out to me yesterday but I thought it was a header for the @ext and @inline js objects.  What kind of things do you picture going inside this if statement in overrides?

techgique: Oh, it is just mean to display the additional `<!-- Conditional -->` HTML comment to distinguish which scripts/styles were conditionally loaded for the given page if one looks at the page source. In case we want to check when debugging a page's JS or CSS, "Does this page have conditional assets?" it's a quick place to look for that. I hadn't intended it to be anything more than that. Did you have other ideas for it in overrides?

jduss4: Ohhhhhhhh that makes sense!  Merging!

PR #12, opened 2017-05-08
Browse Breakdown

techgique:
Split Browse into separate pages; Add flash msgs
Add Orchid-wide SCSS; Tweak nav colors
Remove "project" facet if API URL sets collection

### DIFF
--- a/app/controllers/items_controller.rb
+++ b/app/controllers/items_controller.rb
@@ -1,7 +1,29 @@
 class ItemsController < ApplicationController
 
   def browse
-    @res = $api.query({"facet_num" => 500, "facet_sort" => "term|asc" }).facets
+  end
+
+  def browse_facet
+    # Reverse facet name from url-formatting
+    @browse_facet = params[:facet]
+    if @browse_facet.include?("-")
+      Facets.facet_info.each_with_index do |(facet_name, facet_info), index|
+        if @browse_facet == helpers.datify_name(facet_name)
+          @browse_facet = facet_name
+          break
+        end
+      end
+    end
+
+    # Get selected facet's info
+    @browse_facet_info = Facets.facet_info[@browse_facet]
+    if @browse_facet_info.blank?
+      redirect_to browse_path, notice: "Cannot browse by key: '#{@browse_facet}'"
+      return
+    end
+
+    # Get facet results
+    @res = $api.query({"facet" => @browse_facet, "facet_num" => 500, "facet_sort" => "term|asc" }).facets
   end
 
   def index
--- a/app/controllers/items_controller.rb
+++ b/app/controllers/items_controller.rb
@@ -8,7 +8,7 @@ class ItemsController < ApplicationController
     @browse_facet = params[:facet]
     if @browse_facet.include?("-")
       Facets.facet_info.each_with_index do |(facet_name, facet_info), index|
-        if @browse_facet == helpers.datify_name(facet_name)
+        if @browse_facet == facet_name.parameterize
           @browse_facet = facet_name
           break
         end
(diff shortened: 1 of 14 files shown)


app/controllers/items_controller.rb
```diff
@@ -1,7 +1,29 @@
 class ItemsController < ApplicationController
 
   def browse
-    @res = $api.query({"facet_num" => 500, "facet_sort" => "term|asc" }).facets
+  end
+
+  def browse_facet
+    # Reverse facet name from url-formatting
+    @browse_facet = params[:facet]
+    if @browse_facet.include?("-")
```
jduss4: through the haze of my headcold, I'm having trouble thinking of facets that would have a `-` in their name. Can you give me an example of one?  I can think of `.` for nested fields and ` and date fields?

techgique: The URLs created for `date.year` and `creator.name` with `datify_name` are `/browse/date-year` and `/browse/creator-name`, so if the facet name parameter from the URL includes a dash, I apply the `datify_name` method on each facet's key and compare. If they match, I know that's the facet whose link was clicked.

I think we planned to allow `-` in field names, so I think this will handle any possible mix of facet names containing `-` and `.`, rather than assuming all `-`s in a key from the URL were originally `.`s.

The only mixup I can see is if we had some strange situation where both `abc-abc.abc` and `abc.abc.abc` were both facets. I can't see where that'd happen atm though.

jduss4: 👍

app/views/layouts/body/_contents.html.erb
```diff
@@ -1,3 +1,29 @@
 <div class="container" id="content-wrapper">
+  <% if flash %>
```
jduss4: What are your thoughts on moving all of these to either their own partial / helper, or making it so that a single partial is called multiple times with variables to change it up?  The code for each of the flash notification types looks so similar, it seems like it could be pulled out a bit.  Either way, might be nice to have it in a partial just so that the `_content` isn't full of alert code.

app/views/items/browse.html.erb
```diff
@@ -1,41 +1,22 @@
...
-      </li>
-    <% end %>
-  </ul>
-
-  <!-- content -->
-  <div class="tab-content">
-    <% Facets.facet_info.each_with_index do |(facet_name, facet_info), index| %>
+<h3>Browse by:</h3>
+<div class="row bottom-buffer-lg">
+  <% Facets.facet_info.each_with_index do |(facet_name, facet_info), index| %>
```
jduss4: A couple times now in past projects when we've needed to create rows of a certain number of items, we've used the ruby `each_slice` method to cut it into 3.  I think that could work nicely here to avoid having the modulus at the bottom with a closing / opening div, etc.

Here's an example, albeit with some funky spacing:  https://github.com/CDRH/hopewell_middleman/blob/master/source/catalog_cards.html.erb#L8

app/views/items/browse_facet.html.erb
```diff
@@ -0,0 +1,40 @@
+<h2>Browse by <%= @browse_facet_info["label"] %></h2>
+
+<nav>
+  <ul class="nav nav-tabs bottom-buffer" role="tablist">
+    <% Facets.facet_info.each_with_index do |(facet_name, facet_info), index| %>
+      <li role="presentation" <%= "class=active" if facet_name == @browse_facet %>>
+        <% if facet_name == @browse_facet %>
+          <a href="#"><%= facet_info["label"] %></a>
+        <% else %>
+          <% id = datify_name facet_name %>
```
jduss4: btw, if you have any better ideas of how to do this, I would be so interested.  Once you reminded me how the hyphenated names are working, all the pain of trying to get the facets to work came flooding back to me, haha.

jduss4:
I think it looks good and is quite mergeable, but I left some comments above if you feel like tinkering with it more!

PR #13, opened 2017-05-09
Overrides

jduss4:
I thought this had already been merged...I think cather letters is using it locally right now


app/views/items/show.html.erb
```diff
@@ -1,19 +1,6 @@
...
-    <%= metadata @res, "Works", "works" %>
-    <%= metadata @res, "Place(s)", "places" %>
-    <%= metadata @res, "Annotations", "annotations", false %>
-    <li><strong>TEI XML: </strong><a href="<%= @res['uriXML'] %>"><%= @res["identifier"] %>.xml</a>
-  </ul>
-  <p>
-  <%= @res["text"] %>
-  </p>
-</div>
+<%= render "show_metadata" %>
```
techgique: Would you like to switch all our `render` calls to the abbreviated form?
http://edgeguides.rubyonrails.org/action_view_overview.html#render-without-partial-and-locals-options

I don't recall us ever using the other options, so using the abbreviated version is cool with me.

jduss4: I was using the other version originally because I was passing local variables in (like how the items were being iterated and passed through before this change), but I think they can all be abbreviated now!  Should I go through and change them all over in this PR?

jduss4: Er sorry, not before this change.  Before the change on the highlights PR open, haha.

techgique: It looks like changing the local variable(s) can be abbreviated too, based on the documentation.

Surely! There is a new `render` call in the other PR as well, btw. Just needs the `partial:` string dropped.

jduss4: want to merge that, I'll rebase this on top, then change 'em all?

techgique: You got it!

PR #18, opened 2017-05-16
Adds date range functionality to orchid

jduss4:
Most code copied directly over from Lewis and Clark project
but some changes:
- items_controller date specific code moved to items_helper
- @from and @to cannot be set in helper, so passed to controller
- pulling config date, else random default date being set


jduss4:
@techgique the only breaking change this has for testing is in the config.rb file, which handles setting up the date constants.  I am PRETTY sure that's all that needs to be done, but let me know if it needs anything else to work with the vanilla app!

app/helpers/orchid/date_helper.rb
```diff
@@ -0,0 +1,75 @@
...
+
+  # TODO could abstract this into a CDRH gem
+  # if a date is blank, use the other date in its place
+  def date_overwrite(original, overwrite)
+    return original.reject(&:empty?).blank? ? overwrite : original
+  end
+
+  def date_selection?(from, to)
+    [from, to].each do |date|
+      if !date.blank? && !date.reject(&:empty?).blank?
```
techgique: Switch these to `.present?`?

app/helpers/orchid/date_helper.rb
```diff
@@ -0,0 +1,75 @@
...
+
+    return "#{y}-#{m}-#{d}"
+  end
+
+  # main method responsible for taking params and returning
+  # a set of options for querying the API
+  def date_filter aParams=params
+    options = params.permit!.deep_dup
+    # date search
+    if (!options["date_from"].nil? && !options["date_from"].join('').empty?) \
```
techgique: I think these could be written more succinctly using the trick from below:
```ruby
    if (Array(options["date_from"]).reject(&:empty?).present? \
    ||  Array(options["date_to"]).reject(&:empty?).present?)
```
I tested it locally and it seems to work. Wanna give it a try?

app/helpers/orchid/date_helper.rb
```diff
@@ -0,0 +1,75 @@
...
+    # duplicate parameters, remove date fields, and return
+    options = params.permit!.deep_dup
+    options.delete("date_from")
+    options.delete("date_to")
+    return options
+  end
+
+  # TODO could abstract this into a CDRH gem
+  def date_default(date_params, default_date)
+    y, m, d = date_params.map { |x| x.gsub(/\s+/, "") }
```
jduss4: I honestly don't remember why this got added this way, it's not clear in the original commit from L&C:
https://github.com/CDRH/lewisandclark/commit/ca961ab15cd9c708793a4333599a3570a424e0c9

I'm happy to change it to strip and then if any funny business comes up we can fix it.  Should I apply the fix to L&C, also?

app/helpers/orchid/date_helper.rb
```diff
@@ -0,0 +1,75 @@
...
+    y = default_date[0] if y.blank?
+    m = default_date[1] if m.blank?
+    d = default_date[2] if d.blank?
+
+    # solr can't handle months and days without zero padding (01)
```
jduss4: whoops, missed the solr ref.  Elasticsearch is actually fine without padding, but I kind of like the idea of continuing to do it so that everything is uniform.  I'll change the comment

app/helpers/orchid/date_helper.rb
```diff
@@ -0,0 +1,75 @@
...
+    m = "0#{m}" if m.length == 1
+    d = "0#{d}" if d.length == 1
+
+    return "#{y}-#{m}-#{d}"
+  end
+
+  # main method responsible for taking params and returning
+  # a set of options for querying the API
+  def date_filter aParams=params
```
jduss4: I added it add-hock because I perceived a time when we might want to pass a different object besides the params in, though looking at my code now I realize that `permit!` is going to fail on a normal hash.  Also I realize you're right, I didn't use `aParams`.  I blame my oversight on the sinus headache I've been rocking for several days, haha.

This is the situation I was picturing:
```
# in controller
options = do_something_with_text_query params
options = add_annotations_to_text options
options = date_filter options
```
or something along those lines.  I will wait to change anything to see what you think, but if we don't allow passing something else in, then we would probably need to call the date_filter method first before any other stuff in.  Or we come up with a different system to create the object being passed to the API?

app/helpers/orchid/date_helper.rb
jduss4: Looks good to me.  Do you want me to change it in L&C, too?

https://github.com/CDRH/lewisandclark/commit/162048b06d3d216af6ae94dbb8472db24635c3e4

app/helpers/orchid/date_helper.rb
jduss4: yes

jduss4:
@techgique I went ahead and made some changes despite not talking to you in person, partly because I can't actually speak right now.  Hashtag laryngitis :)
Let me know if you think they look like a good solution

techgique:
Se magnific 👍

PR #24, opened 2017-09-16
Asset pipeline

techgique:
Please test by creating a new Rails app with this version of Orchid. Run the generator and then try adding both global and conditional scripts, stylesheets, and images. If you never see a precompilation error/warning, I'll be able to die in peace. :-D


README.md
```diff
@@ -57,34 +57,93 @@ And then execute:
 bundle install
 ```
 
-## Usage
 
-Once orchid is installed successfully, run a generator to prepare your new rails app. Run this with a `--help` to see what will be changed before you begin. This will NOT erase or overwrite any configuration files, only example configuration files.
+## Usage
+Once Orchid is installed successfully, run the generator to prepare your new rails app. Run this with a `--help` to see which app files will be changed before you begin.
```
jduss4: I realize this was in before your PR, but the `--help` option suggests some arguments that we aren't respecting:

```
Runtime options:
  -f, [--force]                    # Overwrite files that already exist
  -p, [--pretend], [--no-pretend]  # Run but do not make any changes
  -q, [--quiet], [--no-quiet]      # Suppress status output
  -s, [--skip], [--no-skip]        # Skip files that already exist
```

techgique:
Okay, I definitely was in a hurry when I pushed and created my PR on Friday. 😒 

Made some basic fixes as well as a very important change to `application.js` for handling `global/` JS files.

jduss4:
Awesome, it all works for me!

I'm happy to merge it, but first do you think it would be appropriate to add some of the comments that are output to the terminal in this home page view?  Then people could review them more readily as they start working on the app.

image

techgique:
Yeah, I think that's a good idea! Do you think it's good to be in both places so one can read it before they start the app?

jduss4:
Yeah, although it's kind of a pain for us to write it down in two places, I think that it would helpful to have it in both places, or at least a shoutout to look at the README or some centralized file?

techgique:
Yeah duplicating / making more work is the drawback.
We could just have the generator and the index page display a link to the README on GitHub.
How's that sound?

jduss4:
I think that sounds good!  We could mention "for information about updating templates, assets, and more" or something before the link to make it clear

techgique:
Yes, yes. I'll work on it :)

jduss4:
Thanks for moving everything into the readme!  I don't think I realized how involved that would end up being!!

PR #27, opened 2017-09-22
Fix search's "Relevancy" sort link

### LINKED ISSUE #22: When you click away from relevancy and click back you get an error
to reproduce: 

* search for "water"
* click to a sort other than relevancy
* click back to relevancy sort

screen shot 2017-05-18 at 2 05 36 pm

techgique:
Remove sort param for Relevancy link;
sort param value of "score" results in API error

Add comments to explain sort_helper code

Close #22

### DIFF
--- a/app/helpers/orchid/sort_helper.rb
+++ b/app/helpers/orchid/sort_helper.rb
@@ -14,7 +14,16 @@ module Orchid::SortHelper
 
   def sort type, direction="desc"
     new_params = copy_params
-    new_params["sort"] = ["#{type}|#{direction}"]
+
+    # If linking to sort by "score" (relevancy), send no sort[] parameter
+    if type == "score"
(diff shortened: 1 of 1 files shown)


jduss4:
Good find.  I wonder, should we change the API to allow specifically requesting "score" as the sorting criteria as well as having it the default behavior (if no field has been sent in)?  Right now it's defaulting on this line:

https://github.com/CDRH/api/blob/master/app/services/search_item_req.rb#L231

I also have been talking to myself in this issue about sorting.  While making some changes to sorting, I would be happy to add "score" or "relevancy" or something which maps to `_score` in elasticsearch.

techgique:
We're going to plan on "relevancy" being added as a sort type the API will handle. Will implement that in a different PR.

PR #31, opened 2017-09-26
Add section class to <html>; Clear up helper use

### LINKED ISSUE #15: Change class on <html> element
I am not sure if this is still needed after #6, but I'm copying the issue over from the API repo for now.

__Jess__
pages should know their identity and be able to pass controller variable to set class so Karin can use one stylesheet
I recall this was trouble with turbolinks or some other rails caching type of mechanism, specifically with the neihardt version of the site, but I have no idea if Rails 5 can handle it or if other things have changed about how the pages are rendered which would mean this works out better. Should revisit.

Example:
https://github.com/CDRH/neihardt/blob/master/app/views/layouts/application.html.erb#L2

__Karin__

Debating on whether this should always be explicitly set (i.e. passing through a variable from the view) OR if it should be entirely based on the URL. In cocoon I mostly depended on routes/url's. I'd set the home page explicitly:

<html class="home">
and then all the other pages I'd set by the first part of the URL:

www.whatever.org/*/**/***.html
* above = the text that gets inserted as a class name
To be more future proof, we may want to prepend something to make sure that class names don't collide with any framework classes. so maybe

site_home
site_about
site_writings
etc.

Thoughts?

We had a brief discussion about the feasibility of including this based on the url - for most sites, I think this would be enough, though being able to overwrite from a view would also certainly be useful.

__Greg__

Another option may be to use the controller name, params[:controller], similarly to how it is described here: http://edgeguides.rubyonrails.org/asset_pipeline.html#controller-specific-assets

__Jess__

I vaguely recall trying to use the controller previously for the Neihardt test run and it didn't have as fine of control as Karin was looking for. Maybe controller + action?

techgique:
Add class="site_(section)" to <html> element
Application helper site_section called from application layout view
Uses the first value from:
- Manual override in @site_section instance variable
- "home" if a request for the home page
- A sub-uri or namespace name in the path
- Path name if it matches the controller or action name
- "pages" as the last default

Change generator to copy extendable helper files to main app

Remove Orchid::ApplicationHelper include from Orchid::ItemsHelper as
included app-wide by application_helper.rb copied to main app

Close #15

### DIFF
--- a/app/helpers/orchid/application_helper.rb
+++ b/app/helpers/orchid/application_helper.rb
@@ -15,7 +15,7 @@ module Orchid::ApplicationHelper
       # Use override from instance variable
       section = @site_section
     else
-      if current_page?(home_path)
+      if current_page? home_path
         section = "home"
       elsif request.path[/^\/[^\/]+?\/.+/]
         # If request path has a sub-uri section (or namespacing),
--- a/app/helpers/orchid/application_helper.rb
+++ b/app/helpers/orchid/application_helper.rb
@@ -10,4 +10,29 @@ module Orchid::ApplicationHelper
     return new_params
   end
 
+  def site_section
+    if @site_section.present?
+      # Use override from instance variable
+      section = @site_section
+    else
+      if current_page?(home_path)
+        section = "home"
+      elsif request.path[/^\/[^\/]+?\/.+/]
+        # If request path has a sub-uri section (or namespacing),
+        # use that part of the path (e.g. "browse" of "/browse/creator")
+        section = request.path[/^\/([^\/]+?)\/.+/, 1]
+      else
+        # If path matches controller or action name, use that (e.g. "/about")
+        if request.path[/^\/#{params["controller"]}$/]
+          section = params["controller"]
+        elsif request.path[/^\/#{params["action"]}$/]
+          section = params["action"]
+        else
+          section = "pages"
+        end
+      end
+    end
+
+    return " class=\"site_#{section}\"".html_safe
+  end
 end
--- a/lib/generators/setup_generator.rb
+++ b/lib/generators/setup_generator.rb
@@ -28,6 +28,7 @@ class SetupGenerator < Rails::Generators::Base
     msgs << footer_logo
     msgs << gems
     msgs << gitignore
+    msgs << helpers
     msgs << remove_files
     msgs << scripts
     msgs << stylesheet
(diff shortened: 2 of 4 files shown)


app/helpers/orchid/application_helper.rb
```diff
@@ -10,4 +10,29 @@ def clear_search_text
     return new_params
   end
 
+  def site_section
+    if @site_section.present?
+      # Use override from instance variable
+      section = @site_section
+    else
+      if current_page?(home_path)
```
jduss4: since this is a rails method, would it fit the style guide better without parentheses?  I prefer to read it this way, honestly, but I wanted to see if I have the wrong idea in mind for the style guide's preferences.

lib/generators/setup_generator.rb
```diff
@@ -121,6 +122,13 @@ def gitignore
     return "Orchid .gitignore file copied to app's root directory"
   end
 
+  def helpers
+    FileUtils.rm("#{@new_app}/app/helpers/application_helper.rb")
+    FileUtils.cp(Dir.glob("#{@this_app}/app/helpers/*_helper.rb"), "#{@new_app}/app/helpers/")
```
jduss4: I suppose I could test this out, but this is only copying the reference helpers, and not searching recursively the copy the actual filled out helpers, right?

techgique: Correct. As intended. If we copy the actual helpers, then Orchid updates would need to be copied into the main app.

This uses the fancier path searching where it knows to look in engines' helper directories for the include files. Similar to how the `include` directive in the main app's `application.js` will look in the asset directories of engines, gems, vendor/, etc and can grab `orchid.js` from the Gem without copying the file into the main app.

app/helpers/orchid/application_helper.rb
techgique: You're right. I think I wrote it that way because I looked at the API and all their examples are written with parentheses. http://api.rubyonrails.org/classes/ActionView/Helpers/UrlHelper.html#method-i-current_page-3F

I'll fix it :)

lib/generators/setup_generator.rb
jduss4: Right, that's why I wanted to double check that it was only grabbing the referencing helpers.  Cool deal.  :)

app/helpers/orchid/application_helper.rb
jduss4: If you decide you would be okay with switching everything ever to parentheses, let me know, and I will start search and replacing immediately with gusto!

PR #33, opened 2017-10-02
Active? helper

techgique:
Add refactored active? helper; Use with main nav
Add with simpler code and warning if only a regexp passed
Reformat code in _navbar.html.erb to not wrap
Switch other "active" class instances to active?

### DIFF
--- a/app/helpers/orchid/application_helper.rb
+++ b/app/helpers/orchid/application_helper.rb
@@ -1,5 +1,33 @@
 module Orchid::ApplicationHelper
 
+  def active?(value, comparison=nil)
+    if value.class == Regexp
+      if comparison.nil?
+        # Inform if tried to pass only a Regexp to active?
+        raise "current_page? does not support Regexp comparisons"
+      elsif comparison.class == Symbol
+        value.match(params[comparison]) ? "active" : ""
+      else
+        value.match(comparison) ? "active" : ""
+      end
+    else
+      Array(value).each do |v|
+        begin
+          if comparison.nil?
+            return "active" if current_page? v
+          elsif comparison.class == Symbol
+            return "active" if current_page? comparison => v
+          else
+            return "active" if v == comparison
+          end
+        rescue
+          # Rescue when current_page? throws "No route matches" exception
+          return "active" if !comparison.nil? && v == params[comparison]
+        end
+      end
+    end
+  end
+
   def copy_params
     # Remove Rails internal parameters "action" and "controller" from URLs
     # They are always accessible via params["action"] and params["controller"]
@@ -38,4 +66,5 @@ module Orchid::ApplicationHelper
 
     return " class=\"site_#{section}\"".html_safe
   end
+
 end
--- a/app/helpers/orchid/application_helper.rb
+++ b/app/helpers/orchid/application_helper.rb
@@ -1,5 +1,31 @@
 module Orchid::ApplicationHelper
 
+  ##
+  # Return "active" string to class list if path, param, or variable matches
+  #
+  # == Examples
+  #
+  # Request for an explicit path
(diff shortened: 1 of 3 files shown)


app/views/layouts/body/_navbar.html.erb
```diff
@@ -1,18 +1,27 @@
...
     <div class="navbar-collapse collapse">
       <ul class="nav navbar-nav">
-        <li><%= link_to "Home", home_path %></li>
-        <li><%= link_to "Browse", browse_path %></li>
-        <li><%= link_to "Search", search_path %></li>
-        <li><%= link_to "About", about_path %></li>
+        <li class="<%= active?(home_path) %>">
+          <%= link_to "Home", home_path %>
+        </li>
+        <li class="<%= active?(/^\/browse/, request.path) %>">
```
jduss4: Note:  I believe these will break if running the app at a sub-uri (which hasn't happened yet, but something to keep in mind)

app/helpers/orchid/application_helper.rb
```diff
@@ -1,5 +1,33 @@
...
+      else
+        value.match(comparison) ? "active" : ""
+      end
+    else
+      Array(value).each do |v|
+        begin
+          if comparison.nil?
+            return "active" if current_page? v
+          elsif comparison.class == Symbol
+            return "active" if current_page? comparison => v
```
jduss4: While looking up `current_page?`, I found that the api dock documentation uses ( ).  I mention it only because I'm annoyed by the lack of clarity in rails standards.

I had some trouble mentally figuring out what `current_page? comparison => v` was doing, but functionally it is `current_page?({comparison => v})`, right?  Morning brain...

app/helpers/orchid/application_helper.rb
```diff
@@ -1,5 +1,33 @@
 module Orchid::ApplicationHelper
 
+  def active?(value, comparison=nil)
```
jduss4: It would be helpful for me is there is an example or two in a comment above this method!

I'm also trying to figure out how we might write some unit tests for this, because it seems like it would be really awesome to be able to test out if everything is working as we expect.  We would have to figure out how to stub in / mimic / trick `current_page?` to do it, I think?

app/views/layouts/body/_navbar.html.erb
techgique: Good point! I hadn't thought of that. That may be a reason to go with the `:action` comparison instead 🤔

jduss4: :/  our yesterday versions lost the battle of forward thinking

app/helpers/orchid/application_helper.rb
techgique: Okay, I can include a couple examples. I'll probably just copy a couple of the places where it's used in the Orchid code.

Not sure how we'd trick `current_page?`. I think for unit tests we'd almost have to stub in responses. Otherwise it looks like we'd need to use functional testing for regular `current_page?` behavior?

app/helpers/orchid/application_helper.rb
jduss4: Yeah, I realize APIdock isn't the actual rails documentation, but it tends to come up higher in search results so I click it.  :)

app/views/layouts/body/_navbar.html.erb
techgique: Almost

app/helpers/orchid/application_helper.rb
jduss4: This is super helpful, thanks!

techgique: You're welcome! I hope the way I wrote it is clear enough.

PR #38, opened 2017-10-10
Sorting Default & UI Updates

techgique:
Close https://github.com/Willa-Cather-Archive/letters-rails/issues/67

### DIFF
--- a/app/helpers/orchid/sort_helper.rb
+++ b/app/helpers/orchid/sort_helper.rb
@@ -1,15 +1,21 @@
 module Orchid::SortHelper
 
   def sorter
-    current = ""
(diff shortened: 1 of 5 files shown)


app/helpers/orchid/sort_helper.rb
```diff
@@ -1,15 +1,21 @@
...
+      end
     else
-      current = "title|asc"
+      sort_by = params["sort"]
     end
-    render "sort", current: current, query: params["q"].present?
+
+    sort_by = sort_by.first if sort_by.class == Array
+
+    render "sort", sort_by: sort_by
```
jduss4: snazzy

PR #41, opened 2017-10-11
Browse facet sorting

techgique:
Would like both Jess's and Karin's input on the UX/accessibility aspect :)

Close https://github.com/Willa-Cather-Archive/letters-rails/issues/72

### DIFF
--- a/app/assets/stylesheets/orchid.scss
+++ b/app/assets/stylesheets/orchid.scss
@@ -33,6 +33,23 @@
 
 // Site Section-specific
 // =====================
+.site_browse {
+  .browse-item {
+    border-bottom: dotted 0.0125em;
+    margin-bottom: 0.5em;
+    padding-left: 0;
+
+    span {
+      background-color: $light-base;
+      padding: 0 0 0.5em 1em;
+    }
+  }
+
+  a:hover .browse-item {
+    border-bottom: solid 0.0125em;
+  }
+}
+
 .site_search {
   .pagination_container .pagination {
     margin: 0;
--- a/app/controllers/items_controller.rb
+++ b/app/controllers/items_controller.rb
@@ -22,12 +22,12 @@ class ItemsController < ApplicationController
       return
     end
 
-    facet_sort =
+    sort_by =
       params.key?("facet_sort") ? params["facet_sort"] : API_OPTS["facet_sort"]
 
     # Get facet results
     @res = $api.query({"facet" => @browse_facet, "facet_num" => 10000,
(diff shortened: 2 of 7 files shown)


app/assets/stylesheets/orchid.scss
```diff
@@ -33,6 +33,23 @@
 
 // Site Section-specific
 // =====================
+.site_browse {
+  .browse-item {
+    border-bottom: dotted 0.0125em;
```
karindalziel: Is specifying borders in ems a bootstrap convention? It seems like it has the potential for weird results (over pixels). I don't think this should stop a merge, just curious.

techgique: No, just another of my preferences. I write everything in ems. This used to be necessary for relative sizing with zooming but that is a holdover from when browsers wouldn't resize things defined in pixels. Otherwise, I just like to be consistent and decided sticking with ems is easier to figure out than kind of arbitrarily using px in some places and ems in others.

techgique: I think moving some of those sizings into Sass variables could help simplify writing them, but I've gotten used to translating pixel sizes into ems

app/helpers/orchid/sort_helper.rb
```diff
@@ -28,19 +28,29 @@ def sorter
...
+      fields = {
+        "relevancy|desc" => "Relevancy",
+        "rel_separator" => "separator"
+      }
+    else
+      fields = {}
+    end
+
+    # include "grouping" => "separator" to insert separator in dropdown
+    fields.merge({
```
jduss4: I started working on an idea like this and discarded it, though I like how you handled relevancy better.  Could we split the default fields being merged here into their own method so that to override them, you can just plug in a few fields, whereas now to override / alter the fields, you would need to copy this entire method with all the relevancy logic at the topic?

`fields.merge(additional_search_fields)` or something like that, is basically what I'm picturing

app/controllers/items_controller.rb
```diff
@@ -22,9 +22,12 @@ def browse_facet
       return
     end
 
+    facet_sort =
+      params.key?("facet_sort") ? params["facet_sort"] : API_OPTS["facet_sort"]
```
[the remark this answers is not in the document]
techgique: Yeah, I'd considered that but figured it'd only be the case if a user manually emptied the parameter. If it goes through as empty, there's no error at least. In this case.

I can switch it though. Should we maybe just make `param["whatever"].present?` our preferred method for this kind of thing and aim to be consistent in doing that? That seems like a safe default way to code this kind of check.

Issue #39, opened 2017-10-11
"There are a total of 1 item"


jduss4:
We should probably change the text of the search results to be something a little less clunky when there is only one result.

"Your search returned" or "Results returned" ?

PR #42, opened 2017-10-12
Misc patches

jduss4:
- updates orchid generator prompt to use `v1` in demo
- standardizes how pagination is being added to search UI
- right aligns pagination
- pulls search results text into own partial and alters wording

Closes https://github.com/CDRH/orchid/issues/39

### DIFF
--- a/app/assets/stylesheets/orchid.scss
+++ b/app/assets/stylesheets/orchid.scss
@@ -34,6 +34,10 @@
 // Site Section-specific
 // =====================
 .site_search {
+  .pagination_container {
+    text-align:right
+  }
+
   .pagination_container .pagination {
     margin: 0;
   }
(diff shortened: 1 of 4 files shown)


app/assets/stylesheets/orchid.scss
```diff
@@ -34,6 +34,10 @@
 // Site Section-specific
 // =====================
 .site_search {
+  .pagination_container {
```
techgique: The property should be: `text-align: right;` Space between and ending with a semi-colon.

Also use Sass nesting:
```css
.pagination_container {
  text-align: right;

  .pagination {
    margin: 0;
  }
}
```

karindalziel: Do these really need to be nested?

techgique: They don't _need_ to be but I prefer it to DRY up naming the parent class. Plus I like how it mimics the parent-child relationship of the elements being styled. It's visually more like how the markup is nested/indented.

jduss4: it has been done

PR #46, opened 2017-10-13
Switch params checks to `.present?`

techgique:
Also changed `has_key?` found when searching to `key?`

### DIFF
--- a/app/views/items/_search_res_items.html.erb
+++ b/app/views/items/_search_res_items.html.erb
@@ -3,7 +3,7 @@
     <div class="searchResults">
       <div class="row">
         <div class="col-md-12 search_results_info">
-          <% title_display = item.has_key?("title") && item["title"].length > 0 ? item["title"] : "Untitled" %>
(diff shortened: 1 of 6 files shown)


app/views/items/_search_res_items.html.erb
```diff
@@ -3,7 +3,7 @@
     <div class="searchResults">
       <div class="row">
         <div class="col-md-12 search_results_info">
-          <% title_display = item.has_key?("title") && item["title"].length > 0 ? item["title"] : "Untitled" %>
+          <% title_display = item.key?("title") && item["title"].length > 0 ? item["title"] : "Untitled" %>
```
jduss4: If we wanted to make this line shorter, I believe that `.present?` would take care of both if the key existed and if it were an empty string.

techgique: Oh, true. Might as well tidy that up while we're in here. Good catch

PR #47, opened 2017-10-16
fixes bug where relevancy displayed as option when no terms filled in search form

jduss4:


jduss4:
Already addressed by #46, my bad!

PR #51, opened 2017-10-20
Overridable routing

jduss4:
orchid avoid drawing named routes the app is / will be using

creates object to describe routes, iterates through to check for off limit route names


techgique:
@jduss4 will move the documentation to the README and leave a note in routes.rb

jduss4:
I got a little confused about what the two branches route_test and routing_class were being used for, since they looked the same to me, so I put these commits on both branches.  Just FYI

PR #54, opened 2017-10-26
Config

jduss4:
Split the config file into two separate files: private and public.

*This is a breaking change to existing letters-rails apps*
Unlike data repo example, private does not override public; they are distinct files.
I'm not entirely happy with some of the naming conventions / groupings that I came up with
but by and large I tried to put like with like in terms of yaml nestings


lib/generators/templates/private.yml
```diff
@@ -0,0 +1,15 @@
...
+test:
+  <<: *default
+
+development:
+  <<: *default
+  api_path: https://cdrhdev1.unl.edu/api/v1
+
+production:
+  <<: *default
+  api_path: https://cdrhapi.unl.edu
```
techgique: Add `/v1` ?

lib/generators/templates/public.yml
```diff
@@ -0,0 +1,44 @@
...
+    version: 0.1.0
+    # Header <h1>
+    project_name: Sample Template
+    # <title>, meta application-name
+    project_shortname: Sample
+    # Header <h2>
+    project_subtitle: Template Subtitle
+
+  api_options:
+    # only grab letters by default
```
techgique: We should probably change these to not be defaults that fit a specific app, as that will result in headaches with almost every new app. Similarly maybe we should set the default time period to cover from further back until present day or some date in the future so we don't accidentally leave out results by default in new apps?

jduss4: Good idea, I left the filters in but commented out as an example for the future

techgique: Magnifico

PR #56, opened 2017-11-03
Clean up Orchid::Routing to work with Passenger

### LINKED ISSUE #16: Override Orchid Routes in Apps
Right now I don't know how to override or alter routes that were drawn by the orchid engine. I haven't read too much about it, yet, but spitballing ideas

- copy / inject routes into app routes.rb so easily edited there
- figure out some fancy way of manipulating the routes creation (prepend? append?)
- put it under the orchid namespace and then see if it gets called correctly based on location?

http://edgeguides.rubyonrails.org/engines.html#routes

techgique:
Previous approach with `require_relative`, aside from feeling very hacky
and not the Rails way, breaks when deployed via Phusion Passenger
because of its smart spawning method to save resources
and speed up extra app process creation:
https://www.phusionpassenger.com/library/indepth/ruby/spawn_methods/

Symlink `config/routes.rb` from `config/initializers/` to properly load
`ROUTES` constant before main app's routing is processed

Remove class variables as they don't work well with Passenger smart
spawning either. Instead:
- Only call draw method from `routes.rb` if `ROUTES` defined
- Directly check drawn routes to determine whether Orchid should
draw its routes when draw method called

Check whether home path drawn, as apps will always need a home path
and its location should never change, though its controller and view
will often be overridden in the main app

Remove `scope url_prefix` which breaks routes when deployed
via Passenger, our preferred method of deploying Rails apps at sub-URIs

Change `Orchid::Routing` to a module since it is never instantiated

Use `module_function` scoping for draw method per style guide:
https://github.com/bbatsov/ruby-style-guide#module-function

Close #16

### DIFF
--- a/lib/orchid/routing.rb
+++ b/lib/orchid/routing.rb
@@ -13,10 +13,6 @@ module Orchid
           # Add names reserved by main app for more general routes, e.g. '/:id'
           drawn_routes += reserved_names
 
-          # Handle apps with relative url root
-          url_prefix = Rails.application.config.relative_url_root.present? ?
-            Rails.application.config.relative_url_root : "/"
-
           ROUTES.each do |route|
             next if drawn_routes.include?(route[:name])
 
--- a/lib/orchid/routing.rb
+++ b/lib/orchid/routing.rb
@@ -1,41 +1,30 @@
-# Load routes into Orchid::Routing::ROUTES constant before main app's routing
-require_relative "../../config/routes"
-
 module Orchid
-  class Routing
+  module Routing
+    module_function
 
-    # Indicate draw method below available in routes.rb when run by Rails
-    @@ready_to_draw = true
+    def draw(reserved_names: [])
+      # Retrieve list of main app's route names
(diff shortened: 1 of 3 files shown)


lib/orchid/routing.rb
```diff
@@ -1,41 +1,30 @@
...
         Rails.application.routes.draw do
-          # Retrieve list of main app's route names
-          app_route_names = Rails.application.routes.routes.map { |r| r.name }
-
           # Add names reserved by main app for more general routes, e.g. '/:id'
-          app_route_names += reserved_names
+          drawn_routes += reserved_names
 
           # Handle apps with relative url root
           url_prefix = Rails.application.config.relative_url_root.present? ?
```
jduss4: Is url_prefix still being used?

techgique: It's not. Thanks for catching. I'll remove this.

techgique: That project's override requirements are indeed met. It's deployed successfully with Passenger on the dev host.

PR #66, opened 2018-01-05
adds config line to application.rb in generator

jduss4:


lib/generators/setup_generator.rb
```diff
@@ -71,7 +72,7 @@ def copy_configs
     answer = prompt_for_value("Dev API Path", "https://cdrhdev1.unl.edu/api/v1")
     config_replace("private", "api_path: https://cdrhdev1.unl.edu/api/v1", "api_path: #{answer}")
 
-    answer = prompt_for_value("Production API Path", "https://cdrhapi.unl.edu")
+    answer = prompt_for_value("Production API Path", "https://cdrhapi.unl.edu/v1")
```
techgique: Thanks! This had been sticking out like a sore thumb to me when testing new apps recently :)

PR #71, opened 2018-05-21
updates to rails 5.2

jduss4:


techgique:
Looks good so far. Try adding the `bootsnap` gem?

PR #76, opened 2018-07-17
Languages

jduss4:
Pulls basic language components out into locales file

Also addresses
- facet labels / options
- sorting helpers
- date search help text
- documentation
- helper for loading partials by locale


config/locales/en.yml
```diff
@@ -0,0 +1,210 @@
...
+        <p>Searches are not case sensitive. For example: george will come up with the same results as George.</p>
+        <h4>Phrase Search</h4>
+        <p>Searching for a specific phrase may help narrow down the results. Rather long phrases are no problem. For example: "This white pudding we all esteem".</p>
+        <h4>Subsequent Searches</h4>
+        <p>Because of the creative spellings used by the journalists, it may be necessary to try your search multiple times. For example: P?ro*. This search brings up numerous variant spellings of the French word pirogue, "a large dugout canoe or open boat." Searching for P?*r*og?* will bring up other variant spellings. Searching for canoe or boat also may be helpful.</p>
+    filters:
+      # clearing filters (not the text search)
+      clear_all: Clear Filters
+      # the word "filter" singular
+      filter: Filter
```
jduss4: I had a lot of trouble coming up with ways to describe how this is used, since I could see it coming up in more contexts than it is in currently, but I could also put in something more specific about it being for the submit button for date filters?

jduss4: I mean, if we want to be more specific by functionality, I could put it in the search.date section?

techgique: Hmm. It probably can apply to different situations, but I think it'd be even trickier to document multiple places a translation string is used if we tried to generalize them. I think going specific is good with something like "label for the date filters submit button"

config/locales/en.yml
```diff
@@ -1,8 +1,210 @@
...
+      see_more: See more
+      # hidden label to expand a filter category
+      show: "Show %{label} Filters"
+    # placeholder for the text search field
+    placeholder: Search for...
+    results:
+      # how to display number of results when no search / filters have been applied
+      # and therefore user is browsing all items
+      general:
+        zero: 0 results
```
techgique: I'm guessing `pluralize` isn't i18ned? :)

app/helpers/orchid/application_helper.rb
```diff
@@ -75,6 +75,20 @@ def clear_search_text
...
+    params["locale"] || APP_OPTS["language_default"]
+  end
+
+  def localized_partial(partial_name, prefixes)
+    localized = "#{partial_name}_#{locale}"
+    if lookup_context.template_exists?(localized, prefixes, true)
+      "#{prefixes}/#{localized}"
+    else
+      # fallback to informative partial about customization
+      "errors/missing_partial"
```
techgique: Yay! Nice that the partial you render can give the locale in your instructions. I wonder if an `@instance_var` could be used to give the `"#{prefixes}/#{localized}"` string it was looking for too? Could save us a little time in the future? :thinking:

app/views/items/_date_limit.html.erb
```diff
@@ -13,42 +15,43 @@
       <div class="form-group form-inline">
         <% dfrom = params["date_from"] ? params["date_from"] : [] %>
         <div class="clearfix">
-          <%= label_tag "date_from", "Date From" %>
+          <%= label_tag "date_from", t("search.dates.date_from_label", default: "Date From") %>
           <% if date_selection?(params["date_from"], params["date_to"]) %>
-            <%= link_to search_path(clear_dates_params), class: "pull-right btn btn-default btn-xs", role: "button", aria_label: "Clear Date Filter", title: "Clear Date Filter" do %>
+            <%= link_to search_path(clear_dates_params), class: "pull-right btn btn-default btn-xs", role: "button", aria_label: t("search.dates.clear_filter", default: "Clear Date Filter"), title: "Clear Date Filter" do %>
```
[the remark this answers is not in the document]
jduss4: That was a question that I had...I wasn't sure if we should be doing it on the title attribute, since it sounds like screenreaders don't even typically use that?  I dunno, I am happy to go back through and add it if we think it would be helpful, but I ignored it for the purposes of this passthrough.

config/locales/en.yml
```diff
@@ -1,8 +1,210 @@
 en:
+
+  #################
+  # GENERAL PAGES #
+  #################
   general:
+    # footer content with link to CDRH site
+    cdrh_link_html: Created by the <a href="http://cdrh.unl.edu">Center for Digital Research in the Humanities</a>.
```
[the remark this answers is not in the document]
jduss4: good thinking!

config/locales/en.yml
jduss4: Unfortunately, it did not appear to be, but they do have this fancy option picker based on "zero" "one" and "other," plus it got extra tricky for other languages that use different number pluralization types, haha.

For reference:  http://guides.rubyonrails.org/i18n.html#pluralization

app/helpers/orchid/application_helper.rb
jduss4: Interesting idea about the instance variable, I didn't think about that.  I did toy slightly with the notion of passing the current location to the partial as a local variable, although then I didn't just because I figured most of the time the correct partial wouldn't need it.  I do like the idea of leaving ourselves hints for the future, though, so I'll look into that more.

app/views/items/_date_limit.html.erb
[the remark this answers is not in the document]
techgique: The title tag is more for mouse hover so people not using screen readers can still get info about the field, so I think we should i18n that too.

config/locales/en.yml
techgique: Whoa. I like "(paucal)" from the Unicode page on plural forms hehe.
It mentions that the `zero` key is optional, so if that is the same as the plural, we can probably drop it.

jduss4: Is it optional in most of the languages we might be using?  Spanish / German / French?  Since the english one is basically the template I think it would be good to leave it as obvious / encompassing as possible for the sake of the translators who may come.

I realize I'm probably overengineering this since Family Letters might literally be the only project that ever uses this, but I'm paranoid, I guess :)

techgique: Okay. I thought about suggesting just leaving a commented example that uses it but then dropping the redundancy in the uncommented code. I don't know if any of those languages pluralize differently for zero or not, so making it apparent it's available isn't a bad thing.

config/locales/en.yml
```diff
@@ -1,8 +1,210 @@
...
     project_subtitle: Template Subtitle
+
+  #########
+  # ABOUT #
+  #########
+  about:
+    # page title
+    title: About
+    # placeholder "riker ipsum" for about page. If about page has few contents,
+    # it's recommended that you override it in your locales files. If you about
```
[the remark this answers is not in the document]
jduss4: the directory is `config/locales` so I am tempted to just be more specific there

app/views/errors/_missing_partial.html.erb
```diff
@@ -1 +1 @@
-Partial for <%= locale %> was not found, please review your application's customization and add <%= "_filename_#{locale}.html.erb" %> to the appropriate views directory.
+Partial for <%= locale %> was not found, please review your application's customization and add <%= "#{@missing_partial}.html.erb" %>.
```
techgique: This had me confused for a little bit. It's missing the leading underscore `<%= "_#{@missing_partial}.html.erb" %>`

lib/generators/setup_generator.rb
```diff
@@ -71,6 +71,13 @@ def copy_configs
     answer = prompt_for_value("Project Subtitle (Header <h2>)", "Template Subtitle")
     config_set("public", "project_subtitle", answer)
 
+    answer = prompt_for_value("Primary Language", "en")
+    config_set("public", "language_default", answer)
+
+    answer = prompt_for_value("All Languages (separate with a pipe: en|es|de)", "en")
+    puts "ANSWER: #{answer}"
```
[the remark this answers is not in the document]
jduss4: Oh, I think I just forgot to revisit that part after I was trying a few things out.  Originally the config file had a commented line like `# languages: en|es` to demonstrate multiple languages without actually shipping them into the new app that way, so I was trying to select the second instance of it with spaces at the beginning instead of a `#`.  I ended up just putting the `en|es` explanation elsewhere and removing the extra commented line, so I can change this back to no spaces, I believe :)

jduss4:
I think that I'll add some instructions to the top of the facets file so that it's clear to users opening the file what is going on in it!

app/views/layouts/head/_meta.html.erb
```diff
@@ -1,10 +1,24 @@
 <meta name="application-name" content="<%= t "general.project_shortname", default: APP_OPTS["project_shortname"] %>">
-  <% if @description.present? %>
+
```
techgique: The indentation in this file wasn't a mistake. It was intended to keep the indentation in the output (browser view source) consistent, though it's kind of awkward to try to manage that. No need to revert it, just wanted to explain why the file was originally that way.

When it's up to me alone, I just strip out all of the whitespace possible in HTML output where it doesn't affect display in production env. I still leave it there in dev for debugging for invalid HTML like Karin was looking at earlier. Otherwise looking at the layout in the dev tools is nicer anyway. Then I don't see all the awkward indentation resulting from the layers of ERB layouts, views, helpers, and partials. :man_shrugging:

README.md
```diff
@@ -163,7 +163,7 @@ language_default: es
 languages: en|es
 ```
 
-Most of the navigation, buttons, and general wording throughout orchid has been pulled out into `locales/en.yml`.  Copy that file to match the other language(s) your app will support, for example: `locales/es.yml`.  Translate each entry of the yaml file.  You may toggle between languages in the application and view the language differences. **This file must exist for every language your app config specifies, though you may leave it blank.**
+Most of the navigation, buttons, and general wording throughout orchid has been pulled out into `locales/en.yml`.  Copy that file to match the other language(s) your app will support, for example: `locales/es.yml`.  Translate each entry of the yaml file.  You may toggle between languages in the application and view the language differences. **This file must exist for every language your app config specifies, though you may leave it blank.**  Double check that you have changed the key `en` at the top of the file, or else rails will not be able to find the appropriate language text for your application.
```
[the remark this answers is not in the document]
techgique: :thinking: Though the "you may leave it blank" part and needing the key at the top to match is kind of self-contradicting. I'm assuming the file has to exist or there are errors during initialization, whereas the key has to be there when actually switching locale or there are errors at that point. Is that right?

app/models/facets.rb
```diff
@@ -1,3 +1,36 @@
+# ####################
+# Facets Configuration
+# ####################
+#
+# Individual facets must contain the following features:
+#   "[api_field]" => {
+#     "label" => "[label_to_display]",
+#     "display" => [boolean]
+# }
```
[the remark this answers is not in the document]
jduss4: Jess agrees

README.md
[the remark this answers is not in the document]
jduss4: Well, so if you tell it to use `es` and it has no `es` key in a locales file, rails will throw an error saying that language is not allowed.  But if it does find the file, then it's happy enough to start.  At that point, if you had something like `t "search.title", default: "Search"`, and there was no matching `"search.title"` in the `es.yml` file, rails will pull the default "Search" instead.  If you don't put in a default value, then it will display something like "search title."

It's kinda strange, it took me a bit of testing to figure out how all the fallbacks work for the various yaml files.  The process I've figured out is, crudely:

rails app en.yml -> orchid app en.yml -> default value -> the name of the requested translation

app/views/layouts/head/_meta.html.erb
jduss4: Ahh okay. Since I work more with the erb templates I find it far easier to indent there than to think about how it will appear in the source.  Makes sense :)

config/locales/en.yml
```diff
@@ -1,3 +1,10 @@
+# If adding a new language, copy this file, change the "en" code to the new language
+# and translate the text of the fields
+#
+# It is recommended, when adding entirely new fields to this file, to mark in the comment
```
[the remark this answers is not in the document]
jduss4: Good idea, wouldn't be a bad idea to separate them more obviously

app/views/layouts/head/_meta.html.erb
techgique: Yeah, that is much simpler

README.md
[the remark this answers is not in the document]
techgique: I'm just worried about avoiding errors altogether. Would it be simpler to say in the documentation here to at least have the `(lang).yml` _and_ the initial `(lang):` key at the top then, because otherwise it throws the error that it can't find the language?

I guess I don't see the point in telling people they can leave the file blank when that will still lead to another error, just not at app initialization.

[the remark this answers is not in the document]
jduss4: Oh I see, I think I didn't realize I was literally saying to leave the file blank, but rather to leave everything after the key blank.  I'll work it to be more like your wording, since mine is definitely misleading.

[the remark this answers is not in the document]
techgique: Ah, gotcha. Sounds good :+1:

PR #81, opened 2018-07-25
moves title, subtitle, and shortname into locales

jduss4:
also generates locale files based on user input during generation

### DIFF
--- a/lib/generators/setup_generator.rb
+++ b/lib/generators/setup_generator.rb
@@ -23,7 +23,7 @@ class SetupGenerator < Rails::Generators::Base
     msgs = []
     msgs << "\n\nSetup Review\n============"
     msgs << copy_initializer
-    msgs << copy_configs
+    msgs << copy_configs_and_locales
     msgs << facets
     msgs << favicon
     msgs << footer_logo
@@ -31,7 +31,6 @@ class SetupGenerator < Rails::Generators::Base
     msgs << gitignore
     msgs << handle_exceptions
     msgs << helpers
-    msgs << locales
     msgs << remove_files
     msgs << scripts
     msgs << stylesheet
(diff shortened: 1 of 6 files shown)


lib/generators/setup_generator.rb
```diff
@@ -55,35 +54,51 @@ def config_set(config_type, var_name, value)
...
     answer = prompt_for_value("Dev API Path", "https://cdrhdev1.unl.edu/api/v1")
     config_replace("private", "api_path: https://cdrhdev1.unl.edu/api/v1", "api_path: #{answer}")
 
     answer = prompt_for_value("Production API Path", "https://cdrhapi.unl.edu/v1")
     config_replace("private", "api_path: https://cdrhapi.unl.edu/v1", "api_path: #{answer}")
 
-    return "Configuration files copied to config/private.example.yml, config/private.yml, and config/public.yml. Updated with initial app customizations"
+    <<-HEREDOC
+Configuration files copied to config/private.example.yml, config/private.yml, and config/public.yml.
+Locale files copied to config/locales/{lang_default}.yml
```
techgique: Missing the `#` of the string interpolation for `#{lang_default}`?

jduss4: Oh shucks, I didn't mean to turn this into lang_default at all.  It's supposed to be a message for the users like, hey, check the files in locales, but not an actual specific one since it copied allll of them.  Originally it was {language} and then I remember seeing it yesterday and going "language isn't a variable, Jessica!" and changing it to lang_default.  Maybe I should just say `Locale files copied to config/locales` instead.

techgique: Ah okay. That makes sense. I thought that might be a possibility, but since you were using a variable name I really wasn't sure. Your suggestion sounds good :+1:

Issue #68, opened 2018-01-30
rel="nofollow" trials


jduss4:
Look into using `rel="nofollow"` for links on the browse pages and the search filters / sorting. This would prevent search engines / crawlers, from looking through a load of search results pages.

- What do companies / libraries do with their search result links?
- Would this impact screen readers or headless browsers unduly?
- How would the Solr bin/post script handle this?
- Any other concerns?

jduss4:
I asked Michael Fairchild in the MWDC and he believes that the `rel` attributes shouldn't impact screen readers.  Here are supporting links he provided.

From the working draft chapter on links:  https://www.w3.org/TR/html53/links.html#allowed-keywords-and-their-meanings

Accessibility API mappings:  https://www.w3.org/TR/html-aam-1.0/#details-id-245

jduss4:
I don't think that we really need to worry too much about the solr bin/post script because we would only be using that for scraping the website to create a search, but the majority of these links would be on an existing search page!

PR #93, opened 2018-10-24
Nofollow links

jduss4:
Adds rel attributes to describe links
Fixes small bug in display

Closes https://github.com/CDRH/orchid/issues/68

### DIFF
--- a/app/views/items/_paginator.html.erb
+++ b/app/views/items/_paginator.html.erb
@@ -17,7 +17,7 @@
       </li>
     <% else %>
       <li>
-        <%= link_to to_page(current-1, opts) do %>
+        <%= link_to to_page(current-1, opts), rel: "prev" do %>
           <span aria-hidden="true">&laquo;</span>
         <% end %>
       </li>
@@ -26,7 +26,7 @@
     <!-- add the first page if you're not on it, add dots if necessary -->
     <% if current != 1 %>
       <li>
-        <%= link_to "1", to_page(1, opts) %>
+        <%= link_to "1", to_page(1, opts), rel: "nofollow" %>
       </li>
       <% if pages_prior.min != range-1 && current != range-1 %>
         <li class="disabled"><span>...</span></li>
@@ -35,17 +35,17 @@
 
     <!-- prior pages -->
     <% pages_prior.each do |page| %>
-      <li><%= link_to page.to_s, to_page(page, opts) %></li>
+      <li><%= link_to page.to_s, to_page(page, opts), rel: "nofollow" %></li>
     <% end %>
 
     <!-- current page -->
     <li class="active">
-      <%= link_to current.to_s, to_page(current, opts) %>
+      <%= link_to current.to_s, to_page(current, opts), rel: "nofollow" %>
     </li>
 
     <!-- next pages -->
     <% pages_next.each do |page| %>
-      <li><%= link_to page.to_s, to_page(page, opts) %></li>
+      <li><%= link_to page.to_s, to_page(page, opts), rel: "nofollow" %></li>
     <% end %>
 
     <!-- add the last page if you're not on it, add dots if necessary -->
@@ -54,7 +54,7 @@
         <li class="disabled"><span>...</span></li>
       <% end %>
       <li>
-        <%= link_to total.to_s, to_page(total, opts) %>
+        <%= link_to total.to_s, to_page(total, opts), rel: "nofollow" %>
       </li>
     <% end %>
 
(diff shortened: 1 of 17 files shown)


jduss4:
@karindalziel you might want to test this out with your link checker to see if it's accomplishing what we hope!

app/views/items/_paginator.html.erb
```diff
@@ -17,7 +17,7 @@
       </li>
     <% else %>
       <li>
-        <%= link_to to_page(current-1, opts) do %>
+        <%= link_to to_page(current-1, opts), rel: "prev" do %>
           <span aria-hidden="true">&laquo;</span>
```
techgique: Off topic: Are there other attributes we could use here for accessibility's sake? Hiding the link seems less than ideal

jduss4: I dunno!  My knowledge of aria attributes is very poor, but we could open an issue to investigate?

app/views/items/_search_box.html.erb
```diff
@@ -15,10 +15,11 @@
               </span>
             <% end %>
             <%= render "help" %>
-            <%= link_to t("search.view_all", default: "View All Items"), search_path %>
+            <%= link_to t("search.view_all", default: "View All Items"), search_path, rel: "search" %>
           </div>
           <div class="col-md-2">
-            <%= link_to t("search.clear_all", default: "Clear Search"), search_path, :class => "btn btn-danger clear_main_search_text" %>
+            <%= link_to t("search.clear_all", default: "Clear Search"), search_path,
```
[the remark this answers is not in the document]
jduss4: Hmm, that's a good point, I was being overzealous about adding "nofollow" to links :)

jduss4:
@techgique 

1.  link to source data
I don't know if orchid should make too many assumptions, since what comes back from the API could be any path at all.  It seems likely that it will mostly be coming out of cdrhmedia, but I dunno.  I played around for a while with adding attributes about the type to expect, like "it's an alternate of this document and text/xml" but then I didn't want to assume it would always be XML, either, so ultimately I just left it alone.  Maybe I could add a comment in the code above so that when projects override that template, they can include more attributes if they have more information?

2.  I'm kind of wondering if the link_to in the <head> is a mistake, and they should be <link> tags instead?  https://www.w3.org/TR/html52/links.html#link-type-alternate
I just looked up the google recommendations that were one of the places I looked when I was putting in the languages code, and they don't mention putting anything on anchor tags, either, but I don't see how it could hurt?

jduss4:
I'm not ignoring your other comments, I just responded to them directly by commit and apparently they aren't displaying in this PR discussion, now?

app/views/items/_paginator.html.erb
jduss4: Oh wait, I think that the link itself is not hidden, but the << span text is hidden because it's not important for non-visual navigation?

techgique: It doesn't hide the link, but the `aria-hidden` attribute makes it so there is nothing for screen readers to read. The Firefox Dev Tools Accessibility tab seems to confirm this for me. The `aria-hidden` attribute isn't applied if the "«" link is disabled, which is inconsistent too.

I think `aria-label` is what we want. So replace `aria-hidden="true"` with `aria-label="Previous Page"` and `aria-label="Next Page"` respectively. Probably want that with the numbered page links too, e.g. `aria-label="Page 3"`.

techgique:
>  Maybe I could add a comment in the code above so that when projects override that template, they can include more attributes if they have more information?

This seems like a good idea, rather than making an assumption for all Orchid sites :+1: 

>     2. I'm kind of wondering if the link_to in the  is a mistake, and they should be  tags instead?  https://www.w3.org/TR/html52/links.html#link-type-alternate
>        I just looked up the google recommendations that were one of the places I looked when I was putting in the languages code, and they don't mention putting anything on anchor tags, either, but I don't see how it could hurt?

> the link_to in the  is a mistake
In the what? In the meta section?

Yeah, I don't see how it could hurt, but thought I'd ask in case you'd read anything discouraging it. Seems like more relevant semantic information.

app/views/items/_paginator.html.erb
jduss4: Ah, that makes sense.  Agreed that we want to label them all.  Do you want to do that or do you want me to do that?

jduss4:
Yeah, the meta section.  I assume a link to will produce an anchor tag and all the examples I'm seeing for that type of meta stuff are all `<link>` tags (not `<a>` tags), so that's why I'm wondering if I made a mistake there originally?

w3 uses `<link>` and google lists them as well https://support.google.com/webmasters/answer/189077?hl=en, so I think perhaps I'll change those meta tags and add the attributes to the anchor tags as well :)

app/views/items/_paginator.html.erb
techgique: I'm on it with a separate PR

Issue #84, opened 2018-07-30
IIIF image support


jduss4:
Need to write a helper method (and possibly add a setting?) that puts together an image path matching the IIIF format.  Will need to include the collection of a given item.

See SPARC for a previous take on this:
https://github.com/CDRH/sparc/blob/7ed68eed19299338dd24d583258c9b44f354383f/app/helpers/application_helper.rb#L113

jduss4:
#89 might be a good time to work this in

PR #100, opened 2019-01-28
adds iiif image helper support

jduss4:
not backwards compatible -- requires configuration settings
closes https://github.com/CDRH/orchid/issues/84

### DIFF
--- a/migration.md
+++ b/migration.md
@@ -1,5 +1,25 @@
 # Migration
 
+## 2.1.0 to ?
+
+(Future release)
+
+Provides basic support for IIIF image URLs in views with the helper `iiif(partial_image_path, size: "!150,150")`
+
+__config/private.yml__
+
+`iiif_path` must be added, see example in orchid's `private.yml` template
+
+__config/public.yml__
+
+`app_options.media_server_dir` is the name of the project's directory following the IIIF server path
+`app_options.thumbnail_size` is the width and height of the image. Use "!" to preserve the ratio. Ex: "!200,200"
+
+__config/initializers/config.rb__
+
+Add `IIIF_PATH = PRIVATE["iiif_path"]` to this file to make `IIIF_PATH` accessible to application
+
+
 ## 2.0.0 to 2.1.0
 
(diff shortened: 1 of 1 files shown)


jduss4:
Note: I added `size` as a named parameter because I thought there's a possibility that in the future we might also be specifying the rotation, location, color, etc, and so it made sense to include size in such a way that in the future adding more keywords wouldn't mess up older installations

app/helpers/orchid/application_helper.rb
```diff
@@ -75,6 +75,19 @@ def clear_search_text
...
+  #   "documents/doc.0001.jpg" or "doc.1887.82.jpg"
+  # size must be in format "w,h" or "!w,h" like "!200,200"
+  def iiif(image, size: APP_OPTS["thumbnail_size"])
+    server = IIIF_PATH
+    project = APP_OPTS["media_server_dir"]
+    # use %2F for image specific path, / for iiif server path
+    image_esc = image.gsub("/", "%2F")
+    iiif_opts = "full/#{size}/0/default.jpg"
+
+    path = "#{server}/#{project}%2F#{image_esc}/#{iiif_opts}"
```
techgique: Should this be the string as the return value? Is there a need to assign it to `path`?

jduss4: Oops nope, good catch.  I think that might have been a holdover from the sparc methods I was combining

migration.md
```diff
@@ -1,5 +1,25 @@
...
+Provides basic support for IIIF image URLs in views with the helper `iiif(partial_image_path, size: "!150,150")`
+
+__config/private.yml__
+
+`iiif_path` must be added, see example in orchid's `private.yml` template
+
+__config/public.yml__
+
+`app_options.media_server_dir` is the name of the project's directory following the IIIF server path
+`app_options.thumbnail_size` is the width and height of the image. Use "!" to preserve ration. Ex: "!200,200"
```
techgique: Typo `ration` => `ratio`?

Issue #101, opened 2019-01-28
Continue work on facets with no labels


jduss4:
There's an in-progress branch which should be investigated so we can decide if it needs more work or just to be a PR, etc:
https://github.com/CDRH/orchid/tree/missing_labels

jduss4:
The in progress commit mentioned an issue where there were 25 of some item in the "no label" pile, but upon clicking it only 24 showed up in the search results.  I haven't been able to replicate this issue, several months later, so I'm just going to put in a very basic PR to label facets that are an empty string, and not just a nil value.

PR #108, opened 2019-04-16
nil values causing some issues

jduss4:
question:  since this returns `value` by default if no other conditions were met, do we want `value.present?` to return something like "Unknown Format?" or just leave it returning nil like below?

I also tagged on a small issue while checking for empty strings and such related to facets, closes https://github.com/CDRH/orchid/issues/101

### DIFF
--- a/app/helpers/orchid/facet_helper.rb
+++ b/app/helpers/orchid/facet_helper.rb
@@ -84,12 +84,12 @@ module Orchid::FacetHelper
   # in locales yml as person_role, Postal_Card
   def value_label field, value
     info = Facets.facet_info[field]
-    if info && info["translate"]
+    if value.present? && info && info["translate"]
       field_name = field.gsub(".", "_")
       # if this is a list of values, we need to return a list as well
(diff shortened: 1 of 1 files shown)


techgique:
Can you provide an example of how it'll look? I assume we'll want it to stand out somehow so we can clean up any typos etc in the data. If things will intentionally left blank, maybe something like `[blank]`?

@karindalziel may want to weigh in here too

techgique:
Sorry, I had the tab open from yesterday and hadn't seen the new commit. The programming changes look great. Just up to y'all how you think the display of missing labels will be most helpful to us and site viewers. :)

Weirdly it shows this in GitHub on your commit from earlier today:
oct2018

Git time traveling?

jduss4:
@techgique that's from 2018 because that's when I first took a stab at the issue, but had left it sitting there for months!  Decided just to make a PR so we can decide what to do with it, hehe

jduss4:
@techgique right now it looks something like this (it's using the yml locale files for the label)

image

PR #112, opened 2019-04-19
adds to <title> for default pages in orchid

jduss4:
error pages not altered -- will simply display site short name
item show created with overridable method for easier customization
and abbreviated by default if longer than 20 characters

### DIFF
--- a/app/controllers/items_controller.rb
+++ b/app/controllers/items_controller.rb
@@ -1,6 +1,7 @@
 class ItemsController < ApplicationController
 
   def browse
+    @title = t "browse.title"
   end
 
   def browse_facet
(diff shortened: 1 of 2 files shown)


app/controllers/items_controller.rb
```diff
@@ -70,6 +73,19 @@ def show
...
       url = @res["uri_html"]
       @html = Net::HTTP.get(URI.parse(url)) if url
+      @title = item_title
+    end
+  end
+
+  private
+
+  # separated from show action to allow overriding only the
+  # title display for individual items instead of the entire action
```
techgique: :+1:

PR #115, opened 2019-05-07
Api bridge

jduss4:
Merges https://github.com/CDRH/api_bridge gem into Orchid and alters some functionality

Most of changes come after the initial commit of the code importing
- updates syntax to match CDRH preferred
- gets rid of the "helpers" in favor of putting them straight in query class
- redoes a little logic about pagination / parameter encoding / etc
- overrides num with rows for backwards compatibility but adds warning

Also removes stuff that test suite didn't need to run

Have tested this with Nebraskaland but I need to try it with a more intense site to make sure all is well!

closes https://github.com/CDRH/orchid/pull/115


app/services/api_bridge/query.rb
```diff
@@ -33,17 +33,47 @@ class Query
...
+        # use the page and rows to set a start, use default is no num specified
+        num = options.key?("num") ? options["num"].to_i : @options["num"].to_i
+        options["start"] = get_start(page, num)
+      end
+      options
+    end
+
+    # create a version of the request parameters to manipulate
+    # without modifying the original set
+    def clone_and_stringify_options(aOpts)
```
techgique: Note: Resolved by re-written code but I'll leave the note here fyi

I think we ought to avoid using this camel-case letter prefix for variables we're going to clone like this: https://github.com/rubocop-hq/ruby-style-guide#snake-case-symbols-methods-vars

I think `opts_orig` is more clear but also relatively brief. Or just leaving it as `opts` is fine since you have a different name for the derivative object.

test/api_bridge_test.rb
```diff
@@ -0,0 +1,165 @@
...
+    url = @api.create_item_url("cat.let1728")
+    assert_equal "https://test.unl.edu/v1/item/cat.let1728", url
+  end
+
+  test "create_items_url" do
+    # just a facet
+    url = @api.create_items_url({ "facet" => ["date.year", "source"] })
+    assert_equal "https://test.unl.edu/v1/items?facet[]=date.year&facet[]=source", url
+    # a filter that needs some encoding
+    url = @api.create_items_url({ "f" => ["creator.name|\"Yosemite Sam\": Worst Name Ever"]})
```
techgique: The rootin'est, tootin'est test there ever was! :cowboy_hat_face:

test/api_bridge_test.rb
```diff
@@ -0,0 +1,165 @@
...
+
+  test "params_to_hash" do
+    # make sure this hash is not altered and returns only as string keys
+    original = { "a" => 1, b: 2 }
+    cloned = @api.params_to_hash(original)
+    assert_includes cloned, "b"
+    assert_not_includes cloned, :b
+    cloned.delete("a")
+    assert_includes original, "a"
+    # TODO ActionController::Parameters object next!
```
techgique: "TODO" can be removed here now?

test/api_bridge_test.rb
```diff
@@ -0,0 +1,165 @@
...
+  test "prepare_options" do
+    # make sure rows does not override num when also sent in
+    # (@api created with num)
+    opts = @api.prepare_options({ "rows" => 20 })
+    assert_not_includes opts, "rows"
+    assert_equal 5, opts["num"]
+    # TODO figure out if we want rows to override num even though
+    # there is a default num set...will need to figure something out
+    # to determine if num was the default or passed in....??
+    # api = ApiBridge::Query.new("https://test.unl.edu/v1", [], {})
```
techgique: Note: Resolved by decision to always overwrite in newer code but I'll leave the note here fyi

Could set a boolean attribute on `api` when initializing to determine whether num was set with the default or not

test/api_bridge_test.rb
```diff
@@ -0,0 +1,165 @@
...
+  end
+
+  test "query" do
+    skip("skipping integration testing")
+  end
+
+  test "remove_rows" do
+    # TODO see commentary in prepare_options above
+    # we should perhaps prefer rows to num for backwards compatibility
+    # if they are sending it for a specific request
```
techgique: Note: Resolved by decision to always overwrite in newer code but I'll leave the note here fyi

I think this may depend if older sites are setting `rows` or if it's in case a user manually adds it. In the former case, we should probably respect it. In the latter, I'd maybe lean towards ignoring it going forward. Perhaps a deprecation warning in the logs though to let us know someone tried?

test/api_bridge_test.rb
```diff
@@ -0,0 +1,165 @@
...
+  end
+
+  test "set_page" do
+    # no page means page 1!
+    assert_equal 1, @api.set_page(nil)
+    # page can't be zero or negative, sorry
+    assert_equal 1, @api.set_page("0")
+    assert_equal 1, @api.set_page("00")
+    assert_equal 1, @api.set_page("-1")
+    # page has to be some type of number
```
techgique: The code was modified to handle leading zeros so there should probably be a test like:
```ruby
# leading zeros not allowed
assert_equal 1, @api.set_page("09")
```

app/services/api_bridge/query.rb
```diff
@@ -135,10 +136,14 @@ def query(request_options={})
...
     # still supporting for backwards compatibility
+    # overriding num with rows because if an app is specifically
+    # sending a rows request, then we need to override the default
+    # num set when class was initialized
     def remove_rows(opts={})
       if opts.key?("rows")
-        opts["num"] = opts["rows"] if !opts.key?("num")
+        opts["num"] = opts["rows"]
         opts.delete("rows")
+        Rails.logger.warn("Parameter 'rows' is deprecated, please use 'num' instead")
```
techgique: Perhaps add a TODO comment note above this to remove in a future version.

Or if we start needing to deprecate more things we could use a custom `DEPRECATE` note :man_shrugging: : https://guides.rubyonrails.org/command_line.html#notes

test/api_bridge_test.rb
jduss4: 🐴 🤠

app/services/api_bridge/query.rb
jduss4: Yes, this was definitely leftover syntax from my previous programming job.  `aSomething` indicated a parameter being passed in which was then assigned to something else or immediately manipulated into something else.  `_something` was a private method, so it was considered code smell if you ever saw `service.notifier._something()` or the like.  Another noteworthy one that doesn't come up very often here since ruby is structured differently than javascript, was `onSomething()` which indicated a callback / block to execute.

These days I try to only use camel case when I'm programming with JavaScript.  But yeah, that's why I did change this in the modernized-code pass, because our style guide prefers snake case.

test/api_bridge_test.rb
jduss4: yep!

test/api_bridge_test.rb
jduss4: that's true.  since we only have about 3 sites using this it's probably okay, but good idea if we do need to change something fairly major in the future!

test/api_bridge_test.rb
jduss4: Yeah, I know older sites set row on the initialization, which should have been taken care of right at the beginning.  I don't think that any apps we have send query requests with specific rows, though, they all rely on the default behavior set up at the beginning.  So I'm thinking that we should be okay in terms of behavior for our sites, but I can add a log warning.

jduss4: Oh wait, there already was a deprecation warning if rows was set.
https://github.com/CDRH/orchid/pull/115/files#diff-6afeafbfc523a81fe237405f2bfa4e52R146

test/api_bridge_test.rb
jduss4: 👍

test/api_bridge_test.rb
techgique: Sorry, I had seen that but forgot to mention it in the message edit I added after seeing further commits. You were on top of it :smile:

PR #129, opened 2019-06-12
Redirects

jduss4:
Creating a placeholder pull request for the work I've been doing on redirects

Majority of the code was copied from Lewis and Clark code written by @techgique 
but customized to integrate with orchid
added to setup generator
minimal documentation (I don't want to document code that I didn't write because I'm not
entirely sure what all of the options and things might be!)


jduss4:
Handy pages from L&C for reference:
https://github.com/CDRH/lewisandclark/blob/master/lib/cdrh/rewrite.rb
https://github.com/CDRH/lewisandclark/blob/master/config/rewrites.yml
https://github.com/CDRH/lewisandclark/blob/master/config/initializers/rewrite.rb

PR #130, opened 2019-06-13
Conditional Asset Helper and Documentation

techgique:


app/helpers/orchid/application_helper.rb
```diff
@@ -56,6 +56,18 @@ def active?(value, comparison=nil)
     end
   end
 
+  def add_assets(asset_array, *assets)
+    for asset in assets
```
jduss4: I'm kind of curious why you decided to use `for` here instead of just `assets.each` ?  As far as I know, `.each` is the more "ruby" way of doing it.  I am not asking you to change this, because the code works fine and it's very readable, I'm just curious.

In the rubocop styleguide they explain that the biggest difference is that `for x in y` leaves `x` available outside of the `for` statement, whereas `.each do |x|` keeps the scope to the block. See:  https://github.com/rubocop-hq/ruby-style-guide#no-for-loops

techgique: Oops. I just spaced on the `.each` syntax when writing it. I'll change it to match the style guide and Ruby way :)

PR #132, opened 2019-06-18
Add item-specific 404 page, overrideable template

### LINKED ISSUE #131: 404 for item not found pages
If an item is not found in the API, we should send 404 status.  See code changes for Cather here:

https://github.com/Willa-Cather-Archive/letters-rails/pull/291/commits/196d4b9750db9f5675c2e35b50dbf17d285198dd

techgique:
Use i18n strings

Leave image tag commented in template for copying to customize for
specialized 404 pages

Close #131

### DIFF
--- a/app/controllers/items_controller.rb
+++ b/app/controllers/items_controller.rb
@@ -75,6 +75,10 @@ class ItemsController < ApplicationController
       url = @res["uri_html"]
       @html = Net::HTTP.get(URI.parse(url)) if url
       @title = item_title
+    else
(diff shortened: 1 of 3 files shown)


jduss4:
I believe you can remove some code from this file:  https://github.com/CDRH/orchid/blob/dev/app/views/items/show.html.erb

Now that things missing a `@res` are directed to a different template, I believe that the wrapping `if` can be removed from that template along with the current "no item found" code?

PR #133, opened 2019-06-21
Sections

techgique:


app/controllers/items_controller.rb
```diff
@@ -112,5 +113,10 @@ def set_items_api
     end
   end
 
+  def set_page_facets
+    @page_facets = @section.present? ?
```
jduss4: yesssssssss

app/models/orchid/facets.rb
```diff
@@ -34,4 +34,24 @@ def filter_facets
     end
   end
 
+  def section_facets(section)
+    if SECTIONS[section]["facets"].key?(I18n.locale.to_s)
+      facets = SECTIONS[section]["facets"][I18n.locale.to_s]
+    elsif SECTIONS[section]["facets"].key?("en")
```
jduss4: Question to ponder:  we default to english elsewhere if there is no language specified, but I wonder if we should check if there is a default language selected in the app_options, and if so, check for that before defaulting to english?

lib/generators/templates/section.yml
```diff
@@ -16,6 +16,36 @@ default: &default
     # highlighting
     hl_num: 5
 
+  facets:
```
jduss4: I'm picturing that we will probably want to do this for the existing facets also, long term, instead of leaving them in the models file.  Is that what you are picturing, also?

app/models/orchid/facets.rb
techgique: Good idea! I like it

lib/generators/templates/section.yml
techgique: That's definitely where my thoughts were leaning, yes. Where we put them, I was less certain of. Maybe alongside `api_options` in `public.yml` so the organization kind of matches what I did in the section YAML file?

jduss4: I think that makes the most sense to me.  Then the facets.rb file will only be for the methods that delegate which facets to pick and which returning facet values to translate, etc.
Heck, if we don't want that code in the models, I think there's a compelling argument for it going somewhere else, too, I just don't know....where....

techgique: I think we should move it all out of models. The code never interacts with a database, so models seems like the wrong place in hindsight.

I think application_controller.rb is the ideal place since the methods will only be called from application_controller.rb itself and items_controller.rb IIRC.

techgique:
Let me know if you think some of the changes to `render_overridable` will be prone to conflicting with future section templates/partials. If one sets `@section` and has a template or partial with the same name within that section's views directory as some of the templates changed to `render_overridable`, it could become a hassle. I just tried changing all of them and it didn't break anything, so I thought I'd stick with that for now.

app/helpers/orchid/application_helper.rb
```diff
@@ -162,6 +162,55 @@ def prefix_path(path, *args, **kwargs)
     end
   end
 
+  # Render section overrides and localized partials
+  # Partial name does not include the locale, underscore, or extensions
+  # Example:
+  #   render_overridable("explore/partials", "index")
+  #   Looks for in order:
+  #     If @section defined & locale == "es", "@section/_index_es.html.erb"
```
jduss4: thanks for this style of documentation, it is very friendly for how my brain works!

app/helpers/orchid/application_helper.rb
```diff
@@ -162,6 +162,55 @@ def prefix_path(path, *args, **kwargs)
...
+  #     If @section defined & no locale, "@section/_index.html.erb"
+  #     If no @section & locale == "es", "explore/partials/_index_es.html.erb"
+  #     If no @section or locale, "explore/partials/_index.html.erb"
+  #  If no partial found, render an error message with missing partial path
+  def render_overridable(path="", partial="", **kwargs)
```
jduss4: I am kind of confused in this section about how and why path is being assigned / unassigned, I think maybe when you're back if you can talk me through it that would be helpful!

lib/generators/templates/section.yml
techgique: I'm thinking this should be a separate branch/PR as that will definitely break all existing apps. As is, this branch is compatible with existing apps :)

app/helpers/orchid/application_helper.rb
techgique: I took a stab at improving the comments to make this easier to understand. Let me know if this makes sense: https://github.com/CDRH/orchid/pull/133/commits/31ac07612791545428f499861a335c4e8618e81c

It's been a while since we discussed this, so hopefully it will be like coming at this with fresh eyes again and will be a good test for whether it's really clear or not :smile:

app/models/orchid/facets.rb
techgique: Look good?
https://github.com/CDRH/orchid/pull/133/commits/024d56a96fc2d53eaa323d3cc3fc7e0a6971ca24

lib/generators/templates/section.yml
techgique: Going to change app-wide facets to come from a YAML file as well after all :smile:

docs/README.md
```diff
@@ -183,19 +183,93 @@ end
 You can also override an entire controller by simply placing a file with the controller's name in the controllers directory.  For example, `app/controllers/general_controller.rb` would take the place of the Orchid version of this file.  *This approach is not recommended.*
 
 ### Facets
-You may also want to peruse the `app/models/facets.rb` file and alter it for specific fields which you would prefer.  If your app supports multiple languages, you will need to make sure that the facets are organized by language, for example:
+Facets are API fields used to index items in the "Browse" pages. For each
```
jduss4: I find this phrasing a little confusing: `used to index items`, because that makes it sound like they are somehow populating the API to me.  What about something like `Facets are API fields that aggregate results by shared metadata, for example, a "category" listing on the "Browse" page.`  Okay, I don't like that, either....um, any other ideas?

docs/README.md
```diff
@@ -183,19 +183,93 @@ end
...
 
+- `language` - Same as above
+- `api_field` - Same as above except nested field names must be written with
+  underscores replacing periods
+    - `field_value` - The text value returned by the API with underscores
+    replacing periods, commas, and spaces
+    - `translation` - The literal translated text for this particular language
+
+As an example, imagine we have a `source` API field we don't want included in
+search filters or translated, and a `category.en` API field we'd like included
```
jduss4: huh?  Are you imagining that we'll use nested language fields in `category` eventually?  If you're just looking for a nested field to use, might I recommend `person.name` ?  Then you could use it as the field that won't be translated (because the names would (probably) stay the same typically), and then use category as the field that will be translated?

Additionally, did you intend to not have the source field show up at all even on the browse pages for the Spanish version?  That's totally reasonable to show in the example, it just wasn't totally clear from the above text that it was intended to be omitted entirely from browse, not just going untranslated / not on the search page.

Something like

```
facets:
  en:
    category:
      label: Category
      flags:
        - search_filter
    person.name:
      label: "People"
  es:
    category:
      label: Categoría
      flags:
        - search_filter
        - translate
    person.name:
      label: Personas
```

Then maybe a little text as well like, "facet fields do not have to match between languages; one language may utilize 5 fields while another only uses 2"

docs/README.md
techgique: Will push a change after discussing how to avoid confusing language `used to index items` :+1:

docs/README.md
techgique: I'll switch to a completely made up field (which I'll note somehow) `film.title`.

I wasn't intentionally omitting Source in Spanish, but it is a good concept to demonstrate. :sweat_smile: I'll add a comment in the YAML to point out how the languages can use different facets like that. Thanks!

PR #135, opened 2019-09-05
Test prepare options collision

jduss4:
With all the other options, we make it so that if the app setting was "num => 20" and you send in "num => 50", the request will override the app.

on the other hand, some fields like f[] and facet[] are arrays.
app might have something like f[]=collection|whitman set
and if a request sends in f[]=category|writings we don't want
it to completely override the existing array, nor do
we want it to blindly be added to the array if the request
is altering the collection filter
instead, selectively combine the arrays by filtering out any
app wide settings that would conflict with a newly requested one


test/api_bridge_test.rb
```diff
@@ -27,7 +27,7 @@ def before_setup
     facets = ["creator.name", "date.year", "subcategory"]
     assert_equal facets, @api.app_facets
     assert_equal(
-      { "num" => 5, "sort" => "title|asc", "facet" => facets },
+      {"num"=>5, "sort"=>"title|asc", "f"=>["collection|letters"], "facet"=>["creator.name", "date.year", "subcategory"]},
```
techgique: Would you mind restoring the whitespace? :pray: It looks like a really long line that should be split across lines too. Style guide conventions

test/api_bridge_test.rb
```diff
@@ -122,7 +122,7 @@ def before_setup
     req_opts = ActionController::Parameters.new({
       controller: "items", action: "index", utf8: "✓", q: "water"
     })
-    res_opts = { "num"=>5, "sort"=>"title|asc", "facet"=>["creator.name", "date.year", "subcategory"], "q"=>"water", "start"=>0 }
+    res_opts = {"num"=>5, "sort"=>"title|asc", "f"=>["collection|letters"], "facet"=>["creator.name", "date.year", "subcategory"], "q"=>"water", "start"=>0}
```
techgique: Whitespace and split lines again please? :)

test/api_bridge_test.rb
```diff
@@ -27,7 +29,7 @@ def before_setup
     facets = ["creator.name", "date.year", "subcategory"]
     assert_equal facets, @api.app_facets
     assert_equal(
-      {"num"=>5, "sort"=>"title|asc", "f"=>["collection|letters"], "facet"=>["creator.name", "date.year", "subcategory"]},
+      {"num"=>5, "sort"=>"title|asc", "f"=> @f_app, "facet"=>["creator.name", "date.year", "subcategory"]},
```
techgique: Do you want to re-use `@facet_app` here with `"facet" => @facet_app` similar to `"f" => @f_app`?

test/api_bridge_test.rb
```diff
@@ -122,7 +124,7 @@ def before_setup
     req_opts = ActionController::Parameters.new({
       controller: "items", action: "index", utf8: "✓", q: "water"
     })
-    res_opts = {"num"=>5, "sort"=>"title|asc", "f"=>["collection|letters"], "facet"=>["creator.name", "date.year", "subcategory"], "q"=>"water", "start"=>0}
+    res_opts = {"num"=>5, "sort"=>"title|asc", "f"=>["collection|letters", "keyword|production"], "facet"=>["creator.name", "date.year", "subcategory"], "q"=>"water", "start"=>0}
```
techgique: Re-use `@f_app` and `@facet_app` here too?

test/api_bridge_test.rb
```diff
@@ -131,27 +133,32 @@ def before_setup
...
-    # actually returns ["collection|letters", "collection|whitman", "subcategory|marginalia"]
-    setting = ["collection|whitman", "subcategory|marginalia"]
-    assert_equal setting, @api.prepare_options({ "f" => setting })["f"]
+    # if a filter collides (collection in this case), make sure the newly requested one wins out!
+    assert_equal(
+      ["keyword|production", "collection|whitman", "subcategory|marginalia"],
+      @api.prepare_options({ "f" => ["collection|whitman", "subcategory|marginalia"] })["f"]
+    )
+
+    # check that filter collisions where the request has multiple of the same filter preferences request
```
techgique: I think this and the test above are redundant. They are both just testing whether `keyword|production` is removed, right? Whether the replacement has more than one `keyword` or not doesn't matter to the code being tested from how I read it.

app/services/api_bridge/query.rb
```diff
@@ -113,8 +113,21 @@ def prepare_options(options)
       # combine the incoming params with the app default options
       params_hash = params_to_hash(options)
       opts = @app_options.clone.merge(params_hash) { |key, app, req|
-        if app.is_a?(Array) && req.is_a?(Array)
-          app + req
+        if key == "f"
+          # repeated filters are treated as ANDs and for that
```
techgique: This is a little hard to follow, sort of written as run-on sentences.

I suggest something like this
```
# Remove app filter if request contains the same filter.
# Otherwise multiples of a filter are AND-ed in the API query
# and will likely return no results. To preserve any app filters,
# they must be sent along with the request filters.
```

I think you could remove the comment from inside the `reject` block too with this description. I think it covers what we're doing and why pretty well :)

test/api_bridge_test.rb
jduss4: this is partly a casualty of me not being able to boot up git gui today :'(

test/api_bridge_test.rb
jduss4: I mostly wanted to make sure that both of the new keywords would make it through, partly because originally I had written the code in query.rb differently. I think I will remove the more simplistic test rather than this one.

test/api_bridge_test.rb
techgique: :fist_raised: Bad Jess's computer! It's going to have time in the corner later

test/api_bridge_test.rb
```diff
@@ -26,10 +26,14 @@ def before_setup
     assert_equal({ "num" => 50, "facet" => [] }, api.app_options)
 
     # override options
-    facets = ["creator.name", "date.year", "subcategory"]
```
techgique: Good catch. I hadn't noticed the duplicate there!

test/api_bridge_test.rb
```diff
@@ -140,25 +152,36 @@ def before_setup
     assert_equal @f_app, @api.prepare_options({})["f"]
 
     # set a new filter which is not the same as the app wide filter
-    assert_equal ["collection|letters", "keyword|production", "subcategory|memos"], @api.prepare_options({ "f" => ["subcategory|memos"] })["f"]
-
-    # if a filter collides (collection in this case), make sure the newly requested one wins out!
     assert_equal(
-      ["keyword|production", "collection|whitman", "subcategory|marginalia"],
-      @api.prepare_options({ "f" => ["collection|whitman", "subcategory|marginalia"] })["f"]
+      ["collection|letters", "keyword|production", "subcategory|memos"],
```
techgique: I suppose for even more de-duplicating, we could write this as `@f_app + ["subcategory|memos"]`. That mimics the code being tested that concatenates the array contents too. What do you think? :thinking:

jduss4: That's true. I can do either, I don't have a particular preference!

techgique: It'll also simplify rewriting test values if we need to change `@f_app` for future tests. Give it a try?

PR #137, opened 2019-09-05
Reorg Files

techgique:


app/assets/stylesheets/bootstrap-variables.scss
```diff
@@ -51,7 +48,7 @@ $link-hover-decoration: underline !default;
 //
 //## Font, line-height, and color for body text, headings, and more.
 
-$font-family-sans-serif:  "Josefin Sans", "Helvetica Neue", Helvetica, Arial, sans-serif !default;
+$font-family-sans-serif:  Helvetica, Arial, sans-serif !default;
```
karindalziel: We probably still want a better font stack here but since this is just the default maybe it's not a huge deal. I have been stealing the one from bootstrap 4 lately 

-apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, "Helvetica Neue", Arial, "Noto Sans", sans-serif;

even though I don't quite understand it

PR #143, opened 2019-09-23
Sections Improvement

techgique:
Separate reusable routes; Refactor routing code
- Reusable routes moved to a separate array
- Error routes given path helper names
- Error actions reverted to use `render`
- Routing loops and route skipping de-duplicated
- Draw non-reusable routes after others with i18n support
- Revise comments

Simplify prefix and default params for routes
- Refactor routing code to use built-in `scope ..., as: "name"` for
  prefixing path helper names rather than Red Green approach
- Remove extra code from REUSABLE_ROUTES procs
- DRY up routing code even further by consolidating i18n route draws
- Update documentation to remove Red Green approach instructions
    - Rename header to "Section Templates"

### DIFF
--- a/lib/orchid/routing.rb
+++ b/lib/orchid/routing.rb
@@ -7,48 +7,52 @@ module Orchid
       drawn_routes = defined?(Rails.application.routes) ?
         Rails.application.routes.routes.map { |r| r.name } : []
 
-      eval_routes = proc {
-        # if app has specified multiple language support
-        # then they should be included as possible routes
-        # the default language should NOT be specified
-        # as it will not have a locale in the URL
-        langs = APP_OPTS["languages"]
-        if langs.present?
-          locales = Regexp.new(langs)
-          scope "(:locale)", constraints: { locale: locales } do
-            ROUTES.each do |route|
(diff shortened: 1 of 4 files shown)


jduss4:
> prefixing path helper names rather than Red Green approach
like "if the routes don't find you handy they should at least find you handsome" ???

lib/orchid/routing.rb
```diff
@@ -7,48 +7,52 @@ def draw(section: "", routes: [], scope: "")
...
-          scope = "/#{section}" if section.present? && scope.blank?
+      Rails.application.routes.draw do
+        # Set scope to section name if no scope set
+        scope = "/#{section}" if section.present? && scope.blank?
 
-          if scope.present?
-            scope scope do
-              instance_eval(&eval_routes)
+        if scope.present?
+          scope scope do
```
jduss4: `scope scope`, huh?  Could we take this moment to name the variable we are setting to something else so that the code doesn't end up reading like this?

`scope section_scope` or `scope selected_scope` or something along those lines?

jduss4: I realize it was already in the code to begin with, but now that I've noticed it, I'm bothered!

PR #153, opened 2019-10-22
I18n Links

techgique:


app/helpers/orchid/application_helper.rb
```diff
@@ -116,15 +116,24 @@ def locale_link(lang_code)
     url = request.fullpath
 
     if lang_code == APP_OPTS["language_default"]
-      regex = /(?:^\/(?:#{locales}))?(\/.*|$)/
-
-      link_path = url.sub(regex, "\\1")
+      if config.relative_url_root.present?
```
jduss4: Oooh. I was looking for how to get that just the other day and I solved it in a weird and annoying way.....https://github.com/CDRH/family_letters/blob/master/app/controllers/general_override.rb#L11
so I might have to go back and mess around with this relative_url_root thing instead

jduss4:
In some cases on our site, the param "locale" gets caught up in with other stuff, like in search results. In that case, I am having trouble switching between them...

Example URL from localhost:
http://localhost:3000/search?commit=Search+and+Refine&f%5B%5D=language%7Cen&f%5B%5D=creator.name%7CAutobus%2C+Tammy&locale=en&q=hello&sort%5B%5D=relevancy%7Cdesc&utf8=%E2%9C%93

jduss4:
Okay, I figured out how this happened.  If you go to a search page, make the language non-default (en in this case), then select a filter.  Now try to switch back to the default language (es) and you'll notice that things get kinda wonky. I believe it's because the filters themselves are setting the locale incoming, and perhaps that needs to be removed by default when it's like, copying the params or something the way that it probably ditches the action and controller, but I'm not sure.  I'd like to find a more permanent solution with the routes so that something like `link_to {}` on the `/search` will always go to `/search` like it used to, rather than possibly going to `documents/search` like it will currently.

app/helpers/orchid/application_helper.rb
techgique: Thanks Rails folks for making it handy :D

PR #155, opened 2019-10-22
metadata for item "show" multilanguage support

### LINKED ISSUE #152: metadata on item should be multi-language
should be able to rely on same method as search results and facet display
for the item metadata content

Method to use:  https://github.com/CDRH/orchid/blob/dev/app/helpers/orchid/facet_helper.rb#L86

Page that's not doing it:
https://github.com/CDRH/orchid/blob/dev/app/helpers/orchid/display_helper.rb

jduss4:
small potentially breaking change from link: true
change to arguments

closes #152 and removes the need for https://github.com/CDRH/family_letters/commit/0fb03ab330dcc3b1497461d7ac37b00c0219dacf commit

### DIFF
--- a/app/helpers/orchid/display_helper.rb
+++ b/app/helpers/orchid/display_helper.rb
@@ -1,18 +1,18 @@
 module Orchid::DisplayHelper
   # res: API result hash
   # label: field display label ("Category")
-  # search_ele: the corresponding API field
+  # api_field: the corresponding API field
   # link: whether or not text or a link to search results
   #       for the given filter will be displayed, defaults
   #       to displaying a link
-  def metadata(res, label, search_ele, link: true)
-    data = res[search_ele]
+  def metadata(res, label, api_field, link: true)
+    data = res[api_field]
     if data.present?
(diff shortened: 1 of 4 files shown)


app/helpers/orchid/display_helper.rb
```diff
@@ -1,16 +1,21 @@
 module Orchid::DisplayHelper
-  # TODO refactor this?
-  # It will need to handle nested fields if nothing else
-  
-  def metadata(res, label, search_ele, link_bool=true)
+  # res: API result hash
+  # label: field display label ("Category")
+  # search_ele: the corresponding API field
```
techgique: Is "ele" an abbreviation of "element"? Should we maybe change this to `search_field`? That would fit with the description comment better too. `api_field` or just `field` if you want it to be a shorter name :)

jduss4: sounds good to me!  I have a feeling this terminology is like 4 years old, since all this code came out of OSCYS back in the day, and I like the idea of rethinking it.  api_field seems the most straightforward to me!

PR #160, opened 2019-10-24
Make Orchid compatible with Sprockets 4

techgique:
- Hero: https://github.com/rails/sprockets/issues/415#issuecomment-258514974
- Modify default manifest.js to be backwards-compatibile
  with Sprockets 3 defaults and avoid breaking existing apps


jduss4:
Tested with family letters and with adding a new javascript file the Orchid preferred way, and it's looking good to me!

Issue #158, opened 2019-10-23
entering incorrect date causes search to fail


jduss4:
If you enter, say, "1958-13-11" into the search instead of "1958-11-13" because you're not very good with dates, you will get a big ole err saying that there is no `@res` object for things like `@res.count` to use.  Looks like this is because the API is unhappy about the date format (makes sense), so it's returning 500 and then orchid isn't handling it gracefully.

Think about how we might want to modify this behavior, if we want a front-end checker in the UI to refuse to send a request if it doesn't have a valid date, etc.

techgique:
Handling things like this on the back end is preferred priority. I was recently looking at the code that handles removing params etc and I believe the request sent to the API for the date filter has its own method called from the controller.

We might be able to just modify the number down to 12 if it is larger than that. And possibly display a flash message indicating that was done?

As for client-side validation, I think there are built-in markup validation abilities that don't require adding custom JavaScript. Progressive enhancement. Nice browsers get it, lame ones won't.

techgique:
Server-side validation handled, but we should switch the inputs to `<input type="date">` and update associated code for client-side UI and validation

karindalziel:
Verify that this is mostly handled, but we don't need to include in this release.

PR #161, opened 2019-10-24
Server-side date filter input validation

techgique:
Addresses server-side element of #158

### DIFF
--- a/app/helpers/orchid/date_helper.rb
+++ b/app/helpers/orchid/date_helper.rb
@@ -6,7 +6,7 @@ module Orchid::DateHelper
     new_params.delete("date_to")
     # Remove page to return to first page of reorganized results
     new_params.delete("page")
-    return new_params
+    new_params
   end
 
   # TODO could abstract this into a CDRH gem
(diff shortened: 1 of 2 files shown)


app/helpers/orchid/date_helper.rb
```diff
@@ -39,13 +35,34 @@ def date_filter options
...
     options.delete("date_to")
-    return [options, from, to]
+    [options, from, to]
+  end
+
+  def date_format(date)
+    y, m, d = date.split("-")
+
+    # Fix numbers out of range
+    y = 1 if y.to_i < 1
```
jduss4: should this use the year start and year end range that orchid apps specify in the config?  that way if you filled in a bad lewis and clark date, for example, it would start in 1803?

techgique: That was my first inclination, but I hesitated because we won't know if the date being formatted is the first or second date. I wondered if the second date might somehow end up getting set earlier than the first and break things. But I suppose we could add another parameter to the method to indicate that. Sound like a good approach?

jduss4: Ah good point. I think that another parameter makes sense!

techgique: Functionality added. Take a look :)

jduss4: Code looks a lot tighter!  I like it

PR #162, opened 2019-10-24
Gems and Docs

techgique:


README.md
```diff
@@ -1,7 +1,15 @@
 # Orchid
 
-Orchid is a generator which can be used to create a new CDRH API template site.  The new site can either connect to the entire API's contents or filter by "type," meaning that the new site can be used for a specific collection.
+Orchid is a multi-lingual Rails engine which handles common functionality across
```
jduss4: "multi-lingual" is pretty strong for an engine that only has an `en.yml` application :)  Although I've considered that now that we have an `es.yml` translated, maybe we should fold that back into orchid

lib/generators/setup_generator.rb
```diff
@@ -176,13 +176,25 @@ def footer_logo
   end
 
   def gems
-    gem 'bootstrap-sass', '~> 3.3.6'
-    gem 'jquery-rails', '~> 4.3'
+    # Replace sass-rails with sassc-rails
+    gsub_file "#{@new_app}/Gemfile", /^(gem 'sass-rails'.*)$/,
```
jduss4: does this fail silently if there is no sass-rails already in the gemfile?

jduss4: I don't remember how gsub_file behaves if there isn't a match

lib/generators/setup_generator.rb
```diff
@@ -176,13 +176,25 @@ def footer_logo
...
+      "gem 'sassc-rails', '~> 2.1'"
 
     # Remove turbolinks gem
     gsub_file "#{@new_app}/Gemfile", /^(gem 'turbolinks'.*)$/, "#\\1"
 
-    return "Gems: Turbolinks removed"
+    # Replace chromedriver-helper with webdrivers
+    gsub_file "#{@new_app}/Gemfile",
```
jduss4: I think I was picturing that these steps would be taking place in some kind of generator like `update_x_to_x` or whathaveyou.  Most new Rails apps (like if we spun up a new Rails 5 app) wouldn't actually have chromedriver-helper and such things in them anymore, right?  I guess maybe older versions of Rails 5 would?  I'm not sure.

README.md
techgique: If it covers all the facets that are in Orchid, we may as well :)

lib/generators/setup_generator.rb
techgique: I think it would fail silently, but I don't think we'll have to worry about that until Rails 6. I think 5.2.3 still uses `sass-rails` out of the box. Rails 6 will inevitably require a revision of this part of the generator.

lib/generators/setup_generator.rb
techgique: I think 5.2.3 still uses these. That's why I wanted to add this stuff to our generator.

jduss4: okay

README.md
jduss4: I think it should!  Then we'd probably need to write some more logic into the generator to copy that file instead of en.yml, but that can be a patch / minor release later, i reckon

lib/generators/setup_generator.rb
```diff
@@ -119,60 +132,77 @@ def copy_configs_and_locales
...
+                              "https://cdrhdev1.unl.edu/iiif/2")
+    config_replace("private", "iiif_path: https://cdrhdev1.unl.edu/iiif/2",
+                   "iiif_path: #{answer}")
+
+    answer = prompt_for_value("Production IIIF Path",
+                              "https://cdrhmedia.unl.edu/iiif/2")
+    config_replace("private", "iiif_path: https://cdrhmedia.unl.edu/iiif/2",
+                   "iiif_path: #{answer}")
+
+    <<-MSG.chomp
```
jduss4: 👏 I like MSG over HEREDOC, the latter of which means almost nothing to me in terms of instinctual reaction

PR #173, opened 2019-11-25
fixes non-linked translations of metadata

jduss4:
regardless of whether or not something is a link,
we want to display the appropriate language value!


jduss4:
I'm overriding the method in family letters for the launch, but once this is in orchid as a release we should remove the override

Issue #165, opened 2019-10-25
Add Spanish translations


techgique:
From conversation here: https://github.com/CDRH/orchid/pull/162#pullrequestreview-306786175

Add `es.yml` and update generator to copy it to new apps.

PR #180, opened 2019-12-13
Feature/spanish yml

jduss4:
Adding spanish translations from Family Letters to Orchid

Discussion:  Should this be automatically copied into all new applications that specified `es` as a language? I did not implement that for now because if an app specifies `es` this will still work, it will just be more difficult to override, and the next application which uses spanish seems likely far in the future.  I chose not to alter the setup generator because of that rarity, but I did want to add the spanish to Orchid so that changes to the en.yml file will (hopefully) also be modified in the es.yml file as Orchid is developed.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -7,7 +7,6 @@
 #   Ex:  # CDRH - explanation of label
 
 en:
-
   #################
   # GENERAL PAGES #
   #################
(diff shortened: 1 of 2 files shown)


jduss4:
Close if this PR goes through: https://github.com/CDRH/orchid/issues/165

PR #211, opened 2020-10-19
Feature/override titles

jduss4:


app/views/items/_browse_header.html.erb
```diff
@@ -0,0 +1,6 @@
+<h1><%= t "browse.title", default: "Browse" %></h1>
+
+<p>
+  <%= link_to t("browse.all_link", default: "All Items in the Archive"), prefix_path(route_path),
```
techgique, line 4: Should this text be "All Items in the collection"? Maybe this was copied from WCA? :smile:

app/views/items/_browse_facet_header.html.erb
```diff
@@ -0,0 +1 @@
+<h1><%= t "browse.browse_type", default: "Browse by" %> <%= @browse_facet_info["label"] %></h1>
```
techgique, line 1: I think this one should be a lower level heading. Probably an `<h2>`. Because this is the header for the nav on the left side that is followed by the site / section's facets, right? `<h1>` should just be for the overall page titles

techgique (changes requested):
I think using "header" in the partial names is appropriate. They all have an `<h#>` in there so it feels like an appropriate title to me :smile:

app/views/items/_browse_facet_header.html.erb
techgique: It's _not_ the header I thought it was. Nevermind

app/views/items/browse.html.erb
```diff
@@ -1,9 +1,4 @@
-<h1><% t "browse.title", default: "Browse" %></h1>
-
-<p>
-  <%= link_to t("browse.all_link", default: "All Items in the Archive"), prefix_path(route_path),
-        class: "btn btn-lg btn-default", rel: "search" %>
-</p>
+<%= render "browse_header" %>
```
techgique, line 1: I just realized, we probably want these to be `render_overrideable` for easier overrides in sections, right? :man_facepalming:

jduss4: oooooooo

jduss4: good catch, I always forget that :(

Issue #206, opened 2020-08-28
Tell google not to translate language toggle


karindalziel:
On the family letters website, google was trying to translate the "ES/EN" toggle as "IT IS/IN"

I overwrite the template there with the following, but we should change in orchid (then we can remove that template in family letters)

I'm just adding the "notranslate" class to the buttons. 

(this may also be indicative of lang attributes being set incorrectly/not being set somewhere, so experiment to see if we can fix in other ways)

another wayt to fix this may just be to spell out the whole language

```
<% if APP_OPTS["languages"].present? && APP_OPTS["languages"][/\|/] %>
  <div class="language_link btn-group" role="group" aria-label="language choice">
    <% APP_OPTS["languages"].split("|").each do |lang_code| %>
      <%# create link to switch languages that keeps current parameters %>
      <%# TODO a11y / ARIA markup for full language names? %>
      <% if lang_code == I18n.locale.to_s %>
        <%= link_to lang_code.upcase, "#", class: "btn btn-primary disabled notranslate",
          hreflang: lang_code, rel: "alternate", tabindex: -1 %>
      <% else %>
        <%# a11y / ARIA markup for full language name? %>
        <%= link_to lang_code.upcase, locale_link(lang_code),
          class: "btn btn-default notranslate", hreflang: lang_code, rel: "alternate" %>
      <% end %>
    <% end %>
  </div>
<% end %>
```

jduss4:
A couple solutions for this:

One would be to show the language in its own language:  Español / English / 日本語, for example
Another would be to write the language in the current language:  Spanish / English / Japanese, for example

I'm tempted to say we should go for the first one, since after some looking around that appears to be the most common scenario. We could use the yaml files and so if anybody needed to alter them for a specific site, it would not be difficult to change the display.

Issue #154, opened 2019-10-22
i18n language link text


techgique:
We discussed a little and think we probably want the link text for switching locale to display as the language's name in the native form to people who speak it. For example, this is how the Wikipedia home page handles language choice: https://www.wikipedia.org/

jduss4:
Seems like we probably just need to add a locales file to orchid (don't need to copy it into apps, I suspect) that has mappings which would then, presumably, be easy to apply per language in the UI:

```
en:
  en: English
  es: Spanish
  yo: Yoruba
es:
  en: Inglés
  es: Español
  yo: Yoruba
yo:
  en: Gẹẹsi
  es: Sipaki
  yo: Yorùbá
```

PR #215, opened 2020-11-16
Fix/language

jduss4:
Adds full language names to toggle and fixes some exciting stuff along the way!

image

closes https://github.com/CDRH/orchid/issues/206
closes https://github.com/CDRH/orchid/issues/154   (goes with Greg's suggestion, not Jess's bad idea)

### DIFF
--- a/docs/languages.md
+++ b/docs/languages.md
@@ -1,13 +1,20 @@
 # Languages
 
 By default, Orchid assumes you are developing an English-only application, but
-supports the use of multiple languages. The setup generator called with `rails
+supports the use of multiple languages. By default, only English and Spanish
+are supported out of the box, but you may configure any languages you require.
+
+The setup generator called with `rails
 g(enerate) setup` will prompt for your default language and list of all
-languages you would like your application to use.
+languages you would like your application to use. Once generated, you may alter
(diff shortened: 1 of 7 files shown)


docs/languages.md
```diff
@@ -53,11 +60,50 @@ the app can't find a translation of the string.
...
+`&langs` definition, and make it available to your language's locales configuration.
+
+For example, if you were adding Japanese, your
+`config/locales/languages.yml` file would be:
+
+```yaml
+langs: &langs
+  en: English
+  es: Español
+  ja: 日本語
```
techgique, line 83: I think it's more common to abbreviate Japan(ese) as `jp`. `ja` might be Java(nese)

techgique (changes requested):
One tiny change haha. I like this reorganization and the documentation updates :raised_hands:

docs/languages.md
jduss4: Oh interesting.  I was using the LoC list here, although the abbreviation can be arbitrary so it's not like we need to stick to a particular authority list or anything:  https://www.loc.gov/standards/iso639-2/php/code_list.php

techgique: Oh okay. Works for me

PR #217, opened 2020-11-17
fix language config problem

jduss4:
OK there are two issues here:

1.  `languages` appeared in multiple places in the public.yml file, but only the first was intended to be configured based on user language preferences during setup.  To mitigate that, this has now been named `all_languages`

2.  The resulting YAML was still invalid because of recent updates to the psyche gem. Now the thumbnail size is surrounded in quotation marks to make this valid

### DIFF
--- a/app/helpers/orchid/application_helper.rb
+++ b/app/helpers/orchid/application_helper.rb
@@ -182,8 +182,8 @@ module Orchid::ApplicationHelper
   end
 
   def locale_link(lang_code)
-    locales = APP_OPTS["languages"].split("|")
-      .reject { |l| l == APP_OPTS["language_default"] }.join("|")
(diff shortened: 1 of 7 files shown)


app/helpers/orchid/application_helper.rb
```diff
@@ -182,8 +182,8 @@ def locale
   end
 
   def locale_link(lang_code)
-    locales = APP_OPTS["languages"].split("|")
-      .reject { |l| l == APP_OPTS["language_default"] }.join("|")
+    locales = APP_OPTS["all_languages"].split("|")
+      .reject { |l| l == APP_OPTS["all_language_default"] }.join("|")
```
techgique, line 186: I think this was a mistake :persevere:

Issue #187, opened 2020-04-30
namespace generator commands


jduss4:
when I first created the generator I wasn't thinking ahead as much as I should have. Ideally, the `setup` command would actually be `orchid:setup`.  Since apps only run that command on creation, as long as the documentation was updated I don't think such a change would break backwards compatibility?

techgique:
May even want to work on this alongside #181, which would be a very helpful feature

jduss4:
Definitely.  I predict this one will be a real quick fix, whereas the other will require some development, but if we're in there working on stuff we might as well try!

jduss4:
Okay, for now I'm not going to work on #181 at the same time, because that's a rake task and the generator is not, but I will try to get to an update task soon!

PR #219, opened 2020-11-18
Feature/generator

jduss4:
renames `rails g setup` to `rails g orchid_setup` generator task

closes https://github.com/CDRH/orchid/issues/187


lib/generators/setup_generator.rb
```diff
@@ -0,0 +1,9 @@
+class SetupGenerator < Rails::Generators::Base
+
+  desc "Friendly warning that 'setup' is not longer supported"
```
techgique, line 3: `is no longer supported`?

Issue #198, opened 2020-06-01
5xx HTTP status if API response errors


techgique:
Our sites were for all intents and purposes down with the content unavailable from the API due to the intermediate certificate expiration. If we can check that our API responses are failing such that Orchid can tell, it'd be great to return a 5xx HTTP response so site monitoring tools can more easily be made aware something is wrong. I'm not sure if this is possible, but it would be helpful in detecting errors like those seen today.

jduss4:
I will play around with this a little.  If the problem is our API, I think it would be nice to still have the actual site still up, but just have flash warnings all over that say "site is currently experiencing difficulties, not all content available" or something like that.  We could check the API itself if we're doing any kind of site monitoring to see if it's operating.

wkdewey:
I have run into server errors with the API and it appears that the logs and the displayed pages do warn you. I think Jessica fixed this in her PR.

PR #221, opened 2020-11-19
catches api error and adds flash warning to affected pages

jduss4:
possibly addresses https://github.com/CDRH/orchid/issues/198

If the API sends an error response (like for example, if it is down entirely), this will create a friendlier message to the user than the entire page crashing and burning.  We may want to consider, possibly as future issues because I really think this is in general a big improvement, trying to pass through different messages depending on whether there was just a bad query vs that the API is down entirely.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,14 +27,20 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased](https://github.com/CDRH/orchid/compare/v3.1.1...dev) - unreleased
 
+### Fixed
+- displays flash message if API is down, instead of pages throwing errors
+
 ### Changed
 - updates jquery 1.x to 3.x
 - rails g setup is now rails g orchid_setup
+- check_response method added to items_controller
 
 ### Migration
 - modify existing app's `application.js`
   - change `//= require jquery` to `//= require jquery3`
   - apps will continue to function with older jquery if not modified
+  - add `errors.api` and `search.results.error` to locale files
+  - check items_overrides and views/items for changes related to display items when API is unavailable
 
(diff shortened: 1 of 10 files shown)


CHANGELOG.md
```diff
@@ -27,14 +27,20 @@ Versioning](https://semver.org/spec/v2.0.0.html).
...
 ### Changed
 - updates jquery 1.x to 3.x
 - rails g setup is now rails g orchid_setup
+- check_response method added to items_controller
 
 ### Migration
 - modify existing app's `application.js`
   - change `//= require jquery` to `//= require jquery3`
   - apps will continue to function with older jquery if not modified
+  - add `errors.api` and `search.results.error` to locale files
```
techgique, line 42: :raised_hands: These migration notes will be so handy for updating sites with this newer version

app/views/items/_search_res_items.html.erb
```diff
@@ -1,4 +1,4 @@
-<% if @res.count > 0 %>
+<% if @res.count && @res.count > 0 %>
```
techgique, line 1: I think `@res.present?` will work in this case. But the other with `@res.count > 1` wouldn't because that's different logic.

app/views/items/_search_res_count.html.erb
```diff
@@ -1,4 +1,6 @@
-<% if params["q"].blank? && params["f"].blank? %>
+<% if @res.count.nil? %>
```
techgique, line 1: `.blank?` ?

techgique (commented):
Should we change those to use the Rails methods since they're available here?

app/views/items/_search_res_items.html.erb
jduss4: Oh yeah I totally meant to do `@res && @res.count`

PR #223, opened 2021-02-10
slight update to installation instructions

karindalziel:


docs/installation.md
```diff
@@ -14,7 +14,32 @@
...
 
-If you already have Ruby / Rails installed, create a new Rails app (`rails new (your app)`) and add this line to the app's Gemfile:
+Make a new directory with the following files: 
+
+- .ruby-gemset (contents: "[whatever you want to name your gemset]")
+- .ruby-version (contents: "ruby-2.6.6")
+
+cd into the directory again and install ruby if prompted
+
+Next, install the correct verion of ruby (currently the last v5)
```
jduss4, line 24: This should be rails, I believe?

jduss4: What would you think about rephrasing this as `Next, install the most recent version of Rails 5.x ?` or something like that?

karindalziel: @jduss4 I took this suggestion

PR #224, opened 2021-02-11
Fix/favicon

jduss4:
I worked on this last year and figure we should consider where it's at and if we want it in.  I just tested locally and the files are being moved over appropriately by the generator, etc, so that's nice.  It looks like there's a problem leftover from this commit:  https://github.com/CDRH/orchid/commit/e51fd3b9351af5da1d572c01e3d8a61d34414c6e


techgique (approved):
This looks good to me after comparing what's here to the old version and what I see on MDN:
- https://developer.mozilla.org/en-US/docs/Web/HTML/Element/link

They link to a couple generators on this page: https://developer.mozilla.org/en-US/docs/Glossary/Favicon

Were either of these the one y'all used? Just curious for future reference. I didn't try them out to compare and re-open the can of worms about what is best. I trust y'all did your research before.

Only question might be whether the hex colors chosen are what Karin wants.

jduss4:
Looking quickly at the first couple results, I'm guessing it was probably this one?
https://realfavicongenerator.net/

PR #225, opened 2021-02-12
adds role=button to anchor tags which are styled as a button

jduss4:


techgique (approved):
Looks great! I double-checked for any missed elements and looks like you got 'em all :confetti_ball:

techgique:
Oh, could you add a note to the changelog though?

Issue #222, opened 2020-11-23
Expand filters when no JS


jduss4:
`.no-js .collapse { display: block; }`

I think we will need to set a class of `.no-js` on the <body> or <html> or something if the user does not have JS, then we can apply some styles to different site elements that need to be expanded

Issue #125, opened 2019-05-17
Review heading use in templates


techgique:
I used https://validator.w3.org/nu/ to check Cather Letters as a quick test and saw that we're skipping H2's completely. I imagine it's because we're being too influenced by the styling on our headings over their semantic use. I'd like to review our Orchid templates to see if we're doing that anywhere else.

PR #231, opened 2021-12-08
Fix accessibility issues

### LINKED ISSUE #145: Links as Buttons Activation via Spacebar
Buttons are supposed to be able to be activated by spacebar for a11y. I think we should add some JavaScript to site-wide Orchid that selects all `<a class="btn" ...>` elements and makes a spacebar press while focused behave like a click and pressing enter.

In addition, we could add some code that adds `role="button"` if it isn't already present for these elements.

wkdewey:
Fixes #145


wkdewey:
Also fixes #125

wkdewey:
Also fixes #222 . All the js files should be in the right place now and all the manifests should be good.

vendor/assets/javascripts/modernizr-custom.js
```diff
@@ -0,0 +1,255 @@
+/*!
```
techgique, line 1: This is an old version of Modernizr. We should use the latest release: https://github.com/Modernizr/Modernizr/releases

And for actual use by apps, we do want the minified version. Using an unminified version should only be done if debugging something and not committed for use in production.

And could you document the steps used to "build" it so others can update easily in the future? Probably in `/docs/installation.md`

app/assets/stylesheets/orchid.scss
```diff
@@ -121,3 +121,6 @@
     content: " …";
   }
 }
+
+//uncollapse if no javascript
+.no-js .collapse { display: block; }
```
techgique, line 126: Why was this moved to the bottom of the file? I would prefer it where it was before for CSS organization and don't see anything else in our styles that would conflict in terms of CSS precedence here. At least at a glance.

app/assets/config/orchid/manifest.js
```diff
@@ -1,5 +1,6 @@
 // Declare Orchid's assets for auto and pre-compilation
 // which can't be declared with portable relative paths from the generated app
 //= link_tree ../../images/orchid
+//= link_tree ../../images/favicon
```
techgique, line 4: In the future I wouldn't re-commit things that are fixed in another branch to other branches like this. It conflates what branch/PR is handling what and could be error-prone if there are differences between the commits across branches.

If the commits are fixing errors like this, I recommend fixing them but just leaving the fix uncommitted on the branch where other work is being done because they should only be committed for merging on the one other branch.

It wasn't applicable in this case, but if we had gotten the other branch / PR merged before this was re-committed, I would recommend rebasing this branch off of `dev` so the branches merged into `dev` since this branch was started off of it are included as if you had created the branch off the updated `dev` branch.

app/controllers/items_controller.rb
```diff
@@ -84,7 +84,7 @@ def show
 
   # display a flash message if the API response has an error
   def check_response
-    if !@res || (@res.key?("error") && @res.error)
+    if !@res || @res.error
```
techgique, line 87: Same comment here as with what was done with the favicon commit regarding fixes made in other branches. This will actually conflict with what I improved upon the fix in the other branch now because it isn't the same. So you'll have to resolve the conflict and be careful to commit what was done in the other branch rather than this version.

*checks the PR page*

Oh no. Even worse, it's not even flagging it as a conflict. So if I hadn't looked at this closely, this branch would have been merged into `dev` and undone the fixes from the `error_checking` PR :weary:

Honestly, this is probably a good opportunity for you to learn interactive rebasing. That would be the way to make this branch/PR the cleaneast, but we can handle that in January. It would be a good lesson for everyone on the dev team :bulb:

techgique (changes requested):
Some additional work for the new year :partying_face:

app/controllers/items_controller.rb
techgique: Scratch that. It wasn't showing the conflict yet because I hadn't merged the other PR back into `dev`. Now that I did that, it does show the conflict. Still, an interactive rebase would be the best way to remove the work done on the other branches from this one :+1:

wkdewey: I realized this may be an issue, but it's hard to test code when it causes errors. Don't know if it was the case, but what if there were changes in the same file? It would be hard to leave that uncomitted.

wkdewey: maybe I could rebase while I'm still working on the branch?

techgique: An entire file's changes don't have to be committed. I use a git GUI program for line-by-line staging where I want some changes in a file committed and others not. I've never looked into how to do it from the command line, but it must be possible that way too. Just way easier with a GUI.

Now that the other branches have been merged, yeah it is a good time for rebasing. We'll go over it when everyone is back :+1:

app/assets/javascripts/global/button.js
```diff
@@ -0,0 +1,11 @@
+window.addEventListener("load", () => {
+    const buttons = document.querySelectorAll("a[role='button']");
+    buttons.forEach((element) => {
+        element.addEventListener("keyup", (event) => {
+            if (event.keyCode === 32) {
+                event.preventDefault();
+                element.click();
+            }
+        })
```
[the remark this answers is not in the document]
wkdewey: fixed

app/views/items/browse.html.erb
```diff
@@ -1,6 +1,6 @@
 <%= render_overridable "items", "browse_header" %>
 
-<h3><%= t "browse.browse_type", default: "Browse by" %>:</h3>
+<h2><%= t "browse.browse_type", default: "Browse by" %>:</h2>
```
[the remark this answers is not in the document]
wkdewey: fixed

vendor/assets/javascripts/modernizr-custom.js
wkdewey: I have updated it to the latest version and updated the docs.

app/assets/stylesheets/orchid.scss
wkdewey: fixed

app/controllers/items_controller.rb
wkdewey: I attempted a rebase, removing the commits dealing with the favicon assets and the error checking. Hope I didn't mess anything up! It looks like there is no more merge conflict.

techgique: I think the favicon commit is still there but looks like the error checking commit was removed in the rebase. Would still like to go through the process with Karin and Sarita to help them understand how it can be used.

techgique:
Another thing we probably want to tackle with this PR is there are some `<a>` elements that have the `btn` classes but don't have the `role="button"` attribute. We should probably review/standardize that alongside the JS code that leverages that for the keyboard accessibility.

PR #233, opened 2021-12-09
Fix handling of check_response 

### LINKED ISSUE #232: Problem with error checking
```
def check_response
    if !@res || (@res.key?("error") && @res.error)
      flash[:error] = t "errors.api"
    end
  end
```
This code fails if @res is an APIBridge object, with `undefined method `key?' for #<ApiBridge::Response:0x00007ffa05480f10>`.

wkdewey:
This reverts commit e6be8829116bf9436f7a0ad101dc6da3b858ed9d.
Reverts #229 and fixes #232, since the original error-checking code works under most circumstances.

### DIFF
--- a/app/controllers/items_controller.rb
+++ b/app/controllers/items_controller.rb
@@ -39,7 +39,9 @@ class ItemsController < ApplicationController
     }
 
     # Get facet results
-    @res = @items_api.query(options).facets
+    @res = @items_api.query(options)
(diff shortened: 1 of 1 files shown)


wkdewey:
Added some new code to make sure` check_response` is called at the appropriate time when dealing with individual items, in which case `@res` is a hash instead of APIBridge

PR #239, opened 2022-03-10
change metadata lists to use dl and dd

### LINKED ISSUE #213: descriptive term list rather than li for metadata
Currently, our metadata looks like something:

```
<ul>
  <li><strong>Term:</strong> Text goes here</li>
</ul>
```

Perhaps this should instead use a `dl` / `dt` / `dd` solution.  Consider pros and cons and accessibility

wkdewey:
Fixes #213 (I don't know if I got everything)


techgique (changes requested):
The markup looks great, but some accompanying CSS changes are going to be necessary as the elements display as blocks and fill the full width of their container. It _might_ be as simple as changing them to display inline in these contexts, but I'm not certain that will be all that's necessary.
block-display

PR #246, opened 2022-06-02
Bucket aggregations

### LINKED ISSUE #228: faceting person.name by person.role and other nested aggregations
Elsticsearch 6.8 up (at least) has the ability to create a nested aggregation based on another nested value. details here

https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-bucket-nested-aggregation.html

in orchid, this would be very useful to return a list of facets (aggregations) by person.role = attorney

this would necessitate a change to both the API to handle the query and orchid to handle the results (if they are changed) and handle the query setup in public.yml

thinking through this a bit, the api query could look something like this

`facet[]=person.name[by(person.role=attorney)]`

from there, it would probably return a pretty similar list of facets as it currently does, with only a change to the facet return being changing

```
facets:
   person.name:
```

to something like 

```
facets:
    person.name[by(person.role=attorney)]
```

And then in Orchid, we'd have to add the same functionality to detect in public.yml, though it may be handled by the search term in the API

```
      person.name:
        label: People
 ```

becomes

```
    person.name[by(person.role=attorney)]
      label: Attorneys
````

wkdewey:
Will fix #228 but need to test whether it can be generalized to deal with nested and ordinary facets alike.


wkdewey:
I made a copy and rebased it onto dev (since there were some missing changes)

PR #247, opened 2022-06-15
Facet select

### LINKED ISSUE #245: selecting on multiple facets
Do we need to implement this functionality, and is it possible in our current setup? This came up as one of the possible changes to include for Habeas.

wkdewey:
Closes #245


wkdewey:
all commits are in #255 after rebase

PR #254, opened 2022-07-11
Filter display

### LINKED ISSUE #253: Selecting facets with bucket aggregations
With the new modifications, the facets are not showing up at the top, and the selected facet (with the x box to get rid of it) is not showing up if it is from the bucket aggregation.

wkdewey:
Fixes #253


wkdewey:
all the commits are included in #255

PR #255, opened 2022-07-12
release/v4.0.0

### LINKED ISSUE #252: Add option to remove label on metadata
The metadata method (which links the data in the API) automatically puts a label corresponding to the API field. It should be possible to omit this for greater flexibility in layout

wkdewey:
Fixes #252


wkdewey:
I rebased onto Jess's changes, but now I am getting errors so that will need to be worked out

wkdewey:
Specifically, I broke the browse_facet capability for the bucket aggregations, and there are a ton of DEPRECATION WARNINGS (which I should either remove or use a different method).

PR #266, opened 2022-10-31
Deep faceting

### LINKED ISSUE #260: Discuss "deep faceting" implementation
Right now, the solution for "deep faceting" (faceting options that are not in the top 20 or whatever) is to click "more" go to another page, click the facet you want, and then click a search button at the bottom. 

I worry that for longer facet lists, this will be confusing for users because the search button will be hidden, but beyond that, the UI is confusing because the facet options look like links, which would lead a user to think that clicking on one would have an immediate reaction, without the need to click a separate button.

wkdewey:
Fixed #260


app/controllers/items_controller.rb
```diff
@@ -36,7 +36,7 @@ def browse_facet
       params["facet_sort"] : API_OPTS["browse_sort"]
     options = {
       facet: @browse_facet.to_s,
-      facet_num: 10000,
+      facet_limit: 10000,
```
techgique, line 39: Were there no other changes required from changing this? Or is this changed to fit with changes in the API? If so, it should be explained in the commit message. There is no context given for this change so it's kind of a mystery to me as a reviewer.

app/views/items/browse_facet.html.erb
```diff
@@ -20,7 +19,8 @@
     <% end %>
   </ul>
 </nav>
-
+<p>Select facets listed below, then click "search" to see search results</p>
+<%= link_to "Search", prefix_path("home_path", f: params[:f]), rel: "nofollow", class: "btn btn-default" %>
```
techgique, line 23: Again, an explanation in the commit message would help. I think we talked about this in a dev meeting and y'all said it was done so people don't have to scroll all the way to the bottom of a list with thousands of facets, but it should be documented with the code change so it doesn't appear arbitrary to a code reviewer - both when reviewing the PR for merge now and later if we use git history to look back on when / why the code was changed.

techgique (commented):
Let's discuss

PR #269, opened 2022-11-07
Release/v4.0.0

wkdewey:
renaming this branch, also I want to merge the later branches into this one

PR #270, opened 2022-11-21
New release staging

wkdewey:
Replaces https://github.com/CDRH/orchid/pull/269 (`release/4.0.0`). I cherry-picked the last 20 commits.


wkdewey:
The pull requests and branches should be in order now. ‘Release v/4.0.0 has been renamed to `new_release_staging`. My other branches have been rebased or cherry picked on top of that one. These changes resulted in several PR’s being closed (since they were based on the closed branch). I created new PR’s and linked to the old ones. Especially https://github.com/CDRH/orchid/pull/269 which had some commentary. The order is new_release_staging -> rails6 -> rails6_ruby3 -> deep_faceting -> new_release_documentation

On the config code that was confusing: The API expects an array of facets under the key “f”. The config file code reconstructs this array for the new api functionality (a sub array with the parseable original facet name and the alternate name that can be used as an aggregation name). This array gets initialized in app/services/api_bridge/query.rb and put into the options hash for the api query in the `prepare_options` method.  I don't think a hash can substitue for my `[original_facet, aggregation_name]` subarray. I will make more documentation in the code.

wkdewey:
I think this is redundant to #301

PR #275, opened 2022-11-21
Descriptive terms

wkdewey:
See https://github.com/CDRH/orchid/pull/239 which got closed because it was based on a branch I deleted, reopening. I think there were some outstanding issues left on it.

> Fixes https://github.com/CDRH/orchid/issues/213


wkdewey:
Reset the branch to only have the relevant commit

app/helpers/orchid/display_helper.rb
```diff
@@ -36,6 +36,7 @@ def metadata_create_field_link(api_field, item)
     item_label = value_label(api_field, item)
     link_to item_label, prefix_path(route_path, search_params),
       rel: "nofollow"
+    "<dd>#{link}</dd>"
```
techgique, line 39: There's no `link` variable in scope here. The return of `link_to` would have to be assigned to one first. But I think there's another oversight here. This is the only place we have a `<dd>` and `</dd>` tags. This code here only affects output when the `metadata` method isn't called with `link: false`, but it is in many places. So they won't be wrapping non-link metadata info as the non-centralized `<li>` tags in the template were before. I think we actually want the `<dd>` tags in the `.map` above.

app/helpers/orchid/display_helper.rb
```diff
@@ -22,7 +22,7 @@ def metadata(res, label, api_field, link: true, separator: " | ")
         end
       end
       html << dataArray
-                .map { |i| "<span>#{i}</span>" }
+                .map { |i| "#{i}" }
```
techgique, line 25: I think changing this to
```ruby
.map { |i| "<dd>#{i}</dd>" |
```
is what we want to replace _all_ of the old `<li>` tags with `<dd>` tags

techgique (changes requested):
I think I found an oversight in the HTML logic for this

app/helpers/orchid/display_helper.rb
wkdewey: Fixed as suggested

app/helpers/orchid/display_helper.rb
wkdewey: fixed as suggested

wkdewey:
Testing this on whitman, I noticed it caused some styling changes. Maybe that needs to be handled in the specific project rather than Orchid.

Issue #284, opened 2023-01-31
Remove excluded facets from search query


wkdewey:
If facets do not have the search_filter flag, they should not be in the API query for search pages

PR #285, opened 2023-02-08
filter facets from search query if not marked "search filter"

wkdewey:
Fixes https://github.com/CDRH/orchid/issues/284


wkdewey:
all the commits are in #301

PR #340, opened 2025-12-02
Accessibility fixes will

### LINKED ISSUE #319: Search: page title must be updated when the web address changes
Currently, after a user submits search text or uses a filter/facet, the `<title>` of the newly-loaded page is just "Search | {site title}", and the page title (`h1`) is just "Search Results." 

Ideally, this would change to reflect what has been submitted by the user. I think we could at least do this with text input. If the user searches for the text "prairie," for example, when the page reloads, Orchid should make the `<title>` "Search Results" or (even better) "Search Results: 'prairie'" and the `h1` "Search Results: prairie." For faceted/filtered results, it might be more complicated to change the `<title>`, but we could have the `h1` be something like "Search Results: Filtered by Year and Author" or something similar. Not sure.

There's not one right answer on how to do this, but per WCAG 2.1 Level AA guidelines, **the page `<title>` MUST be updated when the web address changes.**

wkdewey:
fixes #319

### DIFF
--- a/app/views/items/_index_header.html.erb
+++ b/app/views/items/_index_header.html.erb
@@ -1,5 +1,9 @@
-<% if params["q"].blank? && params["f"].blank? %>
-  <h1><%= t "search.all_items", default: "All Items" %></h1>
+<% if params["q"].present? && params["f"].present? %>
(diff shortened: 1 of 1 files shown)


erinchambers:
New PR: https://github.com/CDRH/orchid/pull/341

PR #348, opened 2026-03-25
Pagination labels

### LINKED ISSUE #347: Pagination navs each need a unique accessible name
WAE scan result: Landmarks should have a unique role or role/label/title (i.e. accessible name) combination

Failure summary: The landmark must have a unique aria-label, aria-labelledby, or title to make landmarks distinguishable

``` html
<nav aria-label="Pagination">
```
Discovered while auditing PFF https://github.com/CDRH/habeascorpus/issues/416

This is coming from the fact that on Orchid search pages, there is an identical pagination <nav> above and below the search results (_sort_and_page_ui.html.erb, I believe). These two pagination navs need to have unique aria labels even though they're otherwise the same.

erinchambers:
Fixes #347


erinchambers:
Reviewed by @erinchambers & @karindalziel in-person on 2026-03-25

PR #355, opened 2026-07-14
A11y updates for language choice buttons

### LINKED ISSUE #352: Multilingual support for "Clear ___ Filter" buttons and ARIA attributes
Orchid could use some extra multilingual YML for "Clear Filter." Since the Orchid summary buttons now have better aria labels with "Clear ___ Filter" instead of just "Clear Filter," it'd be good to have those words ("Clear" and "Filter") defined in the YML to support translation ("Borrar" and "Filtros" in Spanish, for example).

See Family Letters issue https://github.com/CDRH/family_letters/issues/199

erinchambers:
- Uses semantic `nav` element for Orchid's language choice buttons
- Adds appropriate ARIA attributes to indicate current page
- Adds YAML variables to translate button labels between languages; they will no longer require a Rails view override for this

Also (add 2026-07-14)
- Changes pagination `aria-labelledby` to `aria-label` so that when pagination menu is empty, it is announced as such (instead of h2 and then nothing)

### DIFF
--- a/app/views/layouts/body/_languages.html.erb
+++ b/app/views/layouts/body/_languages.html.erb
@@ -1,9 +1,9 @@
 <% langs = APP_OPTS["all_languages"] %>
 
 <% if langs.present? && langs[/\|/] %>
-  <div class="language_link btn-group" role="group" aria-label="language choice">
+  <nav class="language_link btn-group" role="group" aria-Label="language choice">
     <% langs.split("|").each do |lang_code| %>
       <%= language_toggle_link(lang_code) %>
     <% end %>
-  </div>
+  </nav>
 <% end %>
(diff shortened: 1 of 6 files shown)


app/views/layouts/body/_languages.html.erb
```diff
@@ -1,9 +1,9 @@
 <% langs = APP_OPTS["all_languages"] %>
 
 <% if langs.present? && langs[/\|/] %>
-  <div class="language_link btn-group" role="group" aria-label="language choice">
+  <nav class="language_link btn-group" role="group" aria-Label="language choice">
```
nichgray: should "Label" be capitalized here?

nichgray (approved):
See note about attr capitalization: otherwise this looks good to me!

app/views/layouts/body/_languages.html.erb
erinchambers: It definitely shouldn't - thank you!

